Issue #19, opened 2016-11-05
Implement shooting for player


lethal-guitar:
This breaks down into:

* projectiles (spawning, movement)
* non-solid collision detection (has projectile hit entity?)
* ammo/weapon management

lethal-guitar:
Closing as it's mostly done, and adding individual issues for remaining work.

PR #58, opened 2016-12-22
Add different weapon types for player

lethal-guitar:
Adds support for the different player weapons (laser, flame thrower, rocket launcher). The player's shots now have the correct sound, sprite, and damage multiplier according to the currently carried weapon. The non-regular weapons use ammo, and are reset back to the normal weapon when they are depleted.

Still missing at this point:

- always rapid fire for flame thrower
- laser going through enemies/walls
- flame thrower going through walls?? Not entirely clear yet, but seems like that's the case in the original game
- downwards shot (jetpack) for flame thrower
- weird projectile positioning for left-facing shots as found in the original game

It's also worth noting that the original game doesn't show ammo consumption correctly for the flame thrower. The flame thrower has twice as much ammo (64 instead of 32) as the other weapons. RigelEngine correctly accounts for that, and simply makes the ammo bar go down slower when using the flame thrower. But the original game would use the same scaling as for the other weapons, so the first 32 shots would not show any change on the ammo bar (everything above 32 was simply out of range), and the last 32 would make the bar go down as if the weapon had only ever had 32 shots. I decided that it's not worth it to recreate this bug 😉

### DIFF
--- a/src/data/player_data.hpp
+++ b/src/data/player_data.hpp
@@ -79,6 +79,11 @@ struct PlayerModel {
       mWeapon == WeaponType::FlameThrower ? MAX_AMMO_FLAME_THROWER : MAX_AMMO;
   }
 
+  void switchToWeapon(const WeaponType type) {
+    mWeapon = type;
+    mAmmo = currentMaxAmmo();
+  }
+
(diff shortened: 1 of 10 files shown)


lethal-guitar:
Another step towards #19

PR #328, opened 2019-02-24
Fix blank screen issue

### LINKED ISSUE #326: Black screen, music in background (win10 x64), anyone else?
subj

lethal-guitar:
Fixes #326


pratikone:
Do you know what caused the error ?

lethal-guitar:
@pratikone it was because of uninitialized memory - I assumed the Renderer's `mGlobalTranslation` member was `0, 0`, but due to a change in the glm 3rd party library, it was actually uninitialized, resulting in some random garbage data being interpreted as translation vector, which then caused the rendered image to be moved off screen. The commit message for the 2nd commit has some more details

PR #362, opened 2019-05-29
Clarify getting sources and sub-module also for Windows and OSX in RE…

s-martin:
…ADME.md


lethal-guitar:
Yeah good point, this was kind of buried before. Thanks! 👍

PR #365, opened 2019-05-30
Add a note re macOS unable to unarchive the zip

avioli:
The built-in macOS unarchiver seems to have a hard time unarchiving the shareware `4duke.zip` file download. I used the terminal utility `unzip`, which unarchived it fine. It can be used for the `.SHR` file as well.


lethal-guitar:
Thanks! 👍

Issue #360, opened 2019-05-29
Cannot build on macOS


Kyle-Falconer:
I'm following the instructions in the readme for macOS and I get the following error about a redundant semicolon in glm:

```
[  2%] Built target dbopl
[  6%] Built target entityx
[  8%] Built target glad
[ 10%] Built target speex_resampler
[ 10%] Building CXX object src/CMakeFiles/rigel_core.dir/engine/debugging_system.cpp.o
In file included from /Users/kylefalconer/sources/RigelEngine/src/engine/debugging_system.cpp:17:
In file included from /Users/kylefalconer/sources/RigelEngine/src/engine/debugging_system.hpp:22:
In file included from /Users/kylefalconer/sources/RigelEngine/src/engine/renderer.hpp:24:
In file included from /Users/kylefalconer/sources/RigelEngine/src/engine/shader.hpp:23:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/type_ptr.hpp:37:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../gtc/quaternion.hpp:18:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../gtc/matrix_transform.hpp:24:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../mat4x4.hpp:5:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/.././ext/matrix_double4x4.hpp:5:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../ext/../detail/type_mat4x4.hpp:188:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../ext/../detail/type_mat4x4.inl:1:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../ext/../detail/../matrix.hpp:161:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../detail/func_matrix.inl:1:
In file included from /Users/kylefalconer/sources/RigelEngine/3rd_party/glm/glm/gtc/../ext/../detail/../geometric.hpp:116:
[...]

PR #372, opened 2019-05-30
Make it possible to disable treating warnings as errors

lethal-guitar:
Based on the discussion in https://github.com/lethal-guitar/RigelEngine/issues/360, make it possible to build RigelEngine without treating warnigns as errors. This is to make life easier for those who would like to use RigelEngine, but have no pre-built binaries available for their platform, and who aren't interested in developing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ set(main_exe_name RigelEngine)
 ###############################################################################
 
 option(USE_GL_ES "Use OpenGL ES instead of regular OpenGL" OFF)
+option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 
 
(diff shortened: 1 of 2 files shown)


lethal-guitar:
@AntonioND FYI. I'm hesitant to turn this off by default, because this might lead to surprises for those who would like to actively develop RigelEngine. When you have a branch that builds without errors locally, and then run it on CI (Travis/AppVeyor), you might get a failed build if there are warnings.

I did however add a note about this to the `README`, and changed the instructions to pass the right arguments to CMake in order to disable `-Werror`. So hopefully, for someone on Linux or Mac who follows the command line instructions, the build should be hassle-free even on newer compilers.

AntonioND:
Fair enough!

PR #373, opened 2019-05-31
Fix vcpkg build

### LINKED ISSUE #370: Wrong dependencies, when starting from Windows (using vcpkg)
I'm using `vcpkg` for dependencies under Windows (description see #369).

However, I now have a strange behavior, when I try to start it from VS.

If I run a Debug configuration, `SDL2.dll` is required, but when I run the Release configuration `SDL2d.dll` is required, which is pretty mixed up.

vcpkg ensures that `SDL2d.dll` is in the Debug bin folder and `SDL2.dll` in the Release bin folder. 
All files are actually there, but somehow the Debug program is looking for the Release DLL and the Release program for the Debug dll.

lethal-guitar:
Fixes #370

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,19 +6,16 @@ branches:
     only:
         - master
 
+cache:
+    - c:\tools\vcpkg\installed\
+
 environment:
     BOOST_ROOT: C:\Libraries\boost_1_67_0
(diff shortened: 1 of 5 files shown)


lethal-guitar:
@s-martin this fixed the issues for me, let me if it fixes it for you as well!

s-martin:
Thanks for the quick response. I'll check tomorrow.

s-martin:
Works for me now! 👍

Issue #361, opened 2019-05-29
Cannot build on Windows - VS 2019


s-martin:
I'm trying to build RigelEngine using Visual Studio 2019 16.1.1 under Windows 10 1809.

I get the following compilation errors

> 
> 7>D:\myprojects\RigelEngine\src\engine\shader.cpp(29,69): error C2440:  "Initialisierung": "initializer list" kann nicht in "rigel::engine::GlHandleWrapper" konvertiert werden 
> 7>D:\myprojects\RigelEngine\src\engine\shader.cpp(29,69): message :  Quelltyp konnte von keinem Konstruktor angenommen werden, oder die Überladungsauflösung des Konstruktors ist mehrdeutig
> 7>D:\myprojects\RigelEngine\src\engine\shader.cpp(31,24): error C3536:  "shader": Kann nicht verwendet werden, bevor es initialisiert wurde.
> 7>D:\myprojects\RigelEngine\src\engine\shader.cpp(54,16): error C2440:  "return": "int" kann nicht in "rigel::engine::GlHandleWrapper" konvertiert werden
> 7>D:\myprojects\RigelEngine\src\engine\shader.cpp(54,16): message :  Quelltyp konnte von keinem Konstruktor angenommen werden, oder die Überladungsauflösung des Konstruktors ist mehrdeutig
> 7>D:\myprojects\RigelEngine\src\engine\shader.cpp(65,13): error C2664:  "rigel::engine::GlHandleWrapper::GlHandleWrapper(rigel::engine::GlHandleWrapper &&)" : Konvertierung von Argument 2 von "PFNGLDELETEPROGRAMPROC" in "rigel::engine::GlHandleWrapper::DeleteFunc" nicht möglich

Currently I'm unable to get the reasons for this error.

lethal-guitar:
Huh, weird, I did successfully compile it on VS 2019 on Windows 10, although I might have been using the 2017 toolset, I'd need to check. I'll have a look

s-martin:
I'm using vcpkg to install Boost, SDL2 and SDL2 mixer, but that shouldn't make a difference.

lethal-guitar:
I was still using VS 16.0.4, after I upgraded to 16.1.1, I did get build errors as well. I've opened https://github.com/lethal-guitar/RigelEngine/pull/368 to fix these.

[...]

PR #376, opened 2019-06-02
Updated README with correct instructions to build on Windows

pratikone:
Building on windows with vcpkg requires one to specify the platform type both in vcpkg as well as cmake, otherwise it creates x86 project of Visual Studio. This can alternatively be set in environmental variables. But it is less error-prone this way. Using these commands also automatically finds and uses SDL and Boost libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ For getting the dependencies, I strongly recommend using
 
 
 ```bash
-vcpkg install boost-program-options:x64-windows boost-algorithm:x64-windows sdl2:x64-windows sdl2-mixer:x64-windows
+vcpkg install boost-program-options:x64-windows boost-algorithm:x64-windows sdl2:x64-windows sdl2-mixer:x64-windows --triplet x64-windows
 ```
 
 Then pass `CMAKE_TOOLCHAIN_FILE=C:/path/to/your/vcpkgdir/scripts/buildystems/vcpkg.cmake` when invoking CMake.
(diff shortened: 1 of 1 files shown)


lethal-guitar:
Thanks for the contribution. Did you see https://github.com/lethal-guitar/RigelEngine/issues/361#issuecomment-498008756? Are you sure the `VCPKG_TARGET_TRIPLET` is required? It works fine for me without it. I have only tested with VS 2019 though. Which version of VS, vcpkg and CMake are you using?

Also note that I've updated the README in the meantime (https://github.com/lethal-guitar/RigelEngine/commit/1adb0b54d235e02b09957f336ea7a0c7603831d0) to mention that it's necessary to specify a 64-bit build for Visual Studio 2017. For 2019, this is the default.

pratikone:
Thanks for responding quickly. I had not seen your new responses until now.  Vcpkg is notorious for picking x86 version of packages by default even  if you have both x86 and x64 installed. Infact, in my case it started mixing x64 packages with x64-static ones for compilation. Hence, I suggest that verbosity of command will ensure it works in other systems too. Similarly, for VS 2019, I am using VS 2019 but still cmake generated Vs2017 project for me (maybe because I have vs2017 toolset v141 installed too and it took it by default ). Hence, I suggest that explicit command too.

lethal-guitar:
Ah, I see, so then it was probably working fine for me because I have no other packages installed besides the x64 versions of SDL and Boost? Making sure the commands work properly in different setups sounds good, so once you rebase your PR on `master` I'll be happy to merge.

> Btw, fantastic work on the engine. If you could add some sections about debugging it would add to that. I saw code where you added drawing rectangles for sprites

Thanks! Yeah it's true, it could use some documentation for sure. At the moment, it's not very obvious what you can do. While in game, you can:

* press `B` to toggle showing objects' bounding boxes. objects that harm the player on contact will have their box drawn in red, others in green.
* press `C` to toggle showing collision data. This draws yellow lines on top of the map in all the places where tiles are considered solid. It also colors in ladders and climbable pipes.
* press `G` to toggle drawing a tile grid on top of the map
* press `S` to enter single-step mode. This will stop time from advancing in the game, making everything stop. Pressing space bar advances time by exactly one frame. Pressing `S` again goes back to normal.

The README is already fairly long at this point, so I thought it would be best to add this kind of documentation to the Wiki instead, and then link to it from the README.

lethal-guitar:
@pratikone I just now realized that you fixed a typo as well (`buildystems` -> `buildsystems` in the CMake command example) - thanks again!

Issue #317, opened 2019-01-08
Introduce an enum with all actor IDs


lethal-guitar:
# Background info

The original Duke Nukem II executable uses numeric IDs to refer to specific sprite graphics like enemies etc. The graphics are stored in a way that one such ID can be associated with a group of several related images. For example, the ID `159` refers to the graphic for the "Blue Guard" enemy:

![159_0](https://user-images.githubusercontent.com/1821027/58763887-288a8780-8561-11e9-90c3-5ff4f8fdb929.png)

This enemy has a total of 16 different animation frames, all of which are sub-images of the image group `159`. 

At the same time, the numbers are often also used to identify the corresponding enemy's behavior. If a level file contains an object with the ID `159`, this will cause a Blue Guard to appear in the level.

I have a spreadsheet which lists all the actor IDs occuring in the game along with various information about each one [here](https://docs.google.com/spreadsheets/d/1zI-kjbi3DspsV8EiIySgYhGuXG9VGEhbLdUs71GaBhw/edit?usp=sharing).

# The refactoring

Right now, I have a type alias [`using ActorID = std::uint16_t`](https://github.com/lethal-guitar/RigelEngine/blob/2ad968e6e586ea4b1648dc966bb23dbdbad971c6/src/data/map.hpp#L31) to represent these IDs, and most of the code uses local constants or magic numbers when calling functions that expect an argument of type `ActorID`. The [giant switch statement to configure all actors](https://github.com/lethal-guitar/RigelEngine/blob/2ad968e6e586ea4b1648dc966bb23dbdbad971c6/src/game_logic/entity_configuration.ipp#L698) is also using magic numbers.

To make the code more readable, I think it would be a good idea to change `ActorID` to be an `enum` instead, which gives names to all these ID numbers.

So the work in this refactoring would be to define the `enum`, and then convert existing code to use it. To avoid having to change everything at once, we can maybe temporarily have the `enum` and the type alias co-exist, and then gradually change code until the type alias isn't used anymore and can be deleted.

lethal-guitar:
@s-martin I've put some info in this one now, let me know if you'd like to know more!

PR #377, opened 2019-06-03
Implement enums for ActorID

s-martin:
See #317


lethal-guitar:
@s-martin awesome, thanks for tackling this! I've set up a gitter chat room for the project now: https://gitter.im/RigelEngine/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge feel free to write there if you have any questions.

src/data/map.hpp
```diff
@@ -28,8 +28,311 @@
 
 namespace rigel { namespace data {
 
-using ActorID = std::uint16_t;
+enum class ActorID
```
s-martin: This enum class is a mixture of Actors, FX, etc.
From my point of view a separation doesn't make much sense, but that's to be decided by @lethal-guitar

src/loader/level_loader.cpp
```diff
@@ -306,7 +307,7 @@ LevelData loadLevel(
   for (size_t i=0; i<header.numActorWords/3; ++i) {
     const auto type = levelReader.readU16();
     const base::Vector position{levelReader.readU16(), levelReader.readU16()};
-    actors.emplace_back(LevelData::Actor{position, type, std::nullopt});
+    actors.emplace_back(LevelData::Actor{position, static_cast<ActorID>(type), std::nullopt});
```
s-martin: Not exactly sure, what happens here, if a loaded value is not defined in the enum class.

src/game_logic/entity_configuration.ipp
```diff
@@ -203,13 +203,13 @@ Message MESSAGE_TYPE_BY_INDEX[] = {
...
-    case 213:
-    case 214:
-    case 215:
-    case 216:
-      return id - 213; // 0 to 3
+    case ActorID::Messenger_drone_YOUR_BRAIN_IS_OURS:
+    case ActorID::Messenger_drone_BRING_BACK_THE_BRAIN:
+    case ActorID::Messenger_drone_LIVE_FROM_RIGEL_ITS_SATURDAY_NIGHT:
+    case ActorID::Messenger_drone_DIE:
+      return static_cast<int>(id) - 213; // 0 to 3
```
s-martin: From my point of view the cast is a little bit "hacky", but couldn't think of a better solution so far.

src/game_logic/entity_configuration.ipp
```diff
@@ -335,80 +335,80 @@ ActorID scoreNumberActor(const ScoreNumberType type) {
     int(ScoreNumberType::S5000) == 3 &&
     int(ScoreNumberType::S10000) == 4);
 
-  const auto intType = static_cast<ActorID>(type);
-  return 123 + intType;
+  const auto intType = static_cast<int>(type);
+  return static_cast<ActorID>(static_cast<int>(ActorID::Score_number_FX_100) + intType);
```
s-martin: From my point of view these multiple casts are "hacky", but couldn't think of a better solution so far.

src/loader/level_loader.cpp
```diff
@@ -246,25 +246,26 @@ ActorList preProcessActorDescriptions(
...
-        case 137:
-        case 138:
-        case 142:
-        case 143:
+		case ActorID::Dynamic_geometry_1_falls_down_and_sinks_into_ground:
+		case ActorID::Dynamic_geometry_2_explodes_into_small_parts_flying_away:
+		case ActorID::Dynamic_geometry_3:
+		case ActorID::Dynamic_geometry_4:
+		case ActorID::Dynamic_geometry_5_falls_down_but_remains_intact:
+		// TODO geometry_6??
```
s-martin: @lethal-guitar , not sure, if geometry_6 was left out on purpose.

s-martin:
So far I've implemented all necessary refactorings and it compiles on Windows.

Not sure, how we should handle the `-Werror=switch` errors on Linux/OSX.

src/loader/level_loader.cpp
lethal-guitar: @s-martin nice catch, this is an oversight. Back when I was writing this part of the code, I didn't know that type 6 exists, because none of the levels in the game use it. I discovered it later when reverse-engineering the code for dynamic geometry, but I didn't think of updating this code here.

Feel free to add it for completeness!

src/game_logic/entity_configuration.ipp
lethal-guitar: Agreed, writing out `return 0`, `return 1` etc. for the different cases is probably cleaner and easier to read.

src/loader/level_loader.cpp
lethal-guitar: We could add some verification code that throws if it's not a known actor value. But that's maybe something for a different PR. At the moment, I guess it would store an invalid value in the enum value, and then fail later if t tries to load a Sprite?

src/game_logic/entity_configuration.ipp
lethal-guitar: Yeah agreed, we could do an explicit switch instead? A bit more code, but easier to read?

lethal-guitar:
:tada: Great work, thanks a lot for the contribution! I'll look into the warning on Travis later.

lethal-guitar:
Maybe adding a `default` to the affected switch statements is all that's needed? Probably best to `throw` in there

src/data/map.hpp
lethal-guitar: Was thinking about it - I have some ideas on how it could be separated, but for now,  let's leave it like this :+1:

lethal-guitar:
@s-martin ok, it compiles on Mac and Linux now, too, so it's almost ready to merge! I'm not 100% sure yet if I'll have time, but I'll see if I can get to it this weekend.

lethal-guitar:
@s-martin once again, thank you for this huge contribution! 👍Good job, it's a nice improvement to code readability. I rebased the branch on the latest master, squashed everything down into one commit, and adjusted some formatting here and there.

PR #382, opened 2019-06-12
Fix clamp

### LINKED ISSUE #378: Replace all `base::clamp` with `std::clamp`
When RigelEngine started, it was using C++ 14. `std::clamp` didn't exist yet, so there is a function `base::clamp` [here](https://github.com/lethal-guitar/RigelEngine/blob/1adb0b54d235e02b09957f336ea7a0c7603831d0/src/base/math_tools.hpp#L25).

Now that we can use `std::clamp`, it would be nice to replace all remaining occurrences of `base::clamp` with the former, and then delete `base::clamp`.

s-martin:
fixed #378

### DIFF
--- a/src/base/math_tools.hpp
+++ b/src/base/math_tools.hpp
@@ -22,12 +22,6 @@
 namespace rigel { namespace base {
 
 template<typename T>
-T clamp(const T value, const T minValue, const T maxValue) {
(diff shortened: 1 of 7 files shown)


lethal-guitar:
@s-martin nice one, thank you! 👍 🎉

PR #391, opened 2019-06-27
Refactor upscaling

### LINKED ISSUE #387: Pull up-scaling logic out of `Renderer`
The original Duke Nukem II uses a 320x200 resolution, which on CRTs of the time was stretched vertically to produce a 4:3 aspect ratio image. To keep things simple, RigelEngine currently renders everything into a 320x200 render target, and then scales that up to full screen resolution. This makes things very straightforward, because all the code can be written as if we were still drawing to a screen of that resolution. But this simplicity also means there is not a lot of flexibility. 

Going forward, we'll want to be able to render at full resolution to implement various enhancements. This means up-scaling at the top level needs to be replaced with a system where we can easily switch between full-screen rendering and automatic up-scaling.

There a couple of changes involved in this:

* Make the top-level render target in `game_main.cpp` as big as the full screen resolution, instead of 320x200 (this means we need to pass the window size into `Game` as ctor arguments). Note that we have to keep the top-level render target for fade-in/fade-out effects, even if we're not using it for up-scaling.
* Write a utility function that sets up the global scale, global translation, and clip rect settings of `Renderer` in a way that results in the same up-scaling currently achieved with the 320x200 render target
* Use this utility function to set up up-scaling by default (game modes can override it by changing the scale/translation/clip rect settings accordingly).

lethal-guitar:
Closes #387

### DIFF
--- a/src/base/math_tools.hpp
+++ b/src/base/math_tools.hpp
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <cmath>
+
 
 namespace rigel::base {
 
@@ -34,6 +36,12 @@ auto lerp(const T a, const T b, const float factor) {
 template<typename T>
 bool inRange(const T value, const T min, const T max) {
   return value >= min && value <= max;
+}
+
+
+template<typename T>
+int roundToInt(const T value) {
+  return static_cast<int>(std::round(value));
 }
 
 }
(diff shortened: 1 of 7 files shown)


lethal-guitar:
@pratikone FYI, check it out. I hope this makes things easier for the widescreen support. You should now be able to write a function similar to https://github.com/lethal-guitar/RigelEngine/pull/391/commits/91db522058edfbed8a1e15b4538144134cde4ccf#diff-fd094122cefdc2d5ba5bfcf9cb3c2072R132, and have the full width of the screen usable without things being cut off or stretched.

pratikone:
I tried your this merged PR today and it opens the game in the original 320x240 resolution on top-left corner of the screen. I see that you have `setupSimpleUpscaling()` function but it is not scaling. Is there a way to turn on some scaling?

PR #393, opened 2019-07-01
Fixing the scaling bug which launches the game with default scaling of 1

pratikone:
Even though `setupSimpleUpscaling()` sets up the scaling as widthScale, heightScale, the game still launches with default initial scaling of 1 because game's state is saved even before this new scaling is applied. Fixed it by moving the state saving line to the last.

### DIFF
--- a/src/base/math_tools.hpp
+++ b/src/base/math_tools.hpp
@@ -39,9 +39,15 @@ bool inRange(const T value, const T min, const T max) {
 }
 
 
+template<typename U, typename T>
+U roundTo(const T value) {
+  return static_cast<U>(std::round(value));
+}
+
+
 template<typename T>
-int roundToInt(const T value) {
-  return static_cast<int>(std::round(value));
+int round(const T value) {
+  return roundTo<int>(value);
 }
 
 }
--- a/src/engine/imf_player.cpp
+++ b/src/engine/imf_player.cpp
@@ -16,9 +16,8 @@
 
 #include "imf_player.hpp"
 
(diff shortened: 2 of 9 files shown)


lethal-guitar:
That's odd - this shouldn't be necessary :thinking:  The idea is that the state saver captures the state from before the new scaling etc. is configured, gets returned from `setupSimpleUpscaling` via move or copy elision, and is then captured in a variable (enforced via `nodiscard`). Only when that variable goes out of scope should settings be reset to what they were at the time of the state saver's construction.

It's possible that this doesn't quite work on older compilers, which compiler version are you using? I'm using VS 2019, but I had to add a workaround to make it compile on CI for VS 2017. I never tried building and running with VS 2017 locally, so it's possible that there's some issue there.

Most likely the move ctor for `StateSaver` doesn't do the right thing, now that I think about it, and I guess VS2017 doesn't elide the copy in this case. I can check it out sometime later today!

lethal-guitar:
@pratikone I didn't have much time today to test this, so it might not be the final fix, but maybe you can give this a shot: https://gist.github.com/lethal-guitar/2bf419bd71b49288db5105f15e24dea7. Does this fix the up-scaling for you, even when you undo the change in this PR?

At least in theory, it should make returning the state saver from a function work correctly regardless of copy elision.

lethal-guitar:
Hopefully fixed by https://github.com/lethal-guitar/RigelEngine/pull/394

pratikone:
Closing pull request as referenced #394  fixes the issue.

PR #394, opened 2019-07-02
Implement correct move-semantics for Renderer::StateSaver

lethal-guitar:
Without this, functions constructing a StateSaver and returning it have to rely on copy elision in order to work properly, but that's not guaranteed to happen on all compilers in all situations.

Most likely, this is responsible for the up-scaling not working for you @pratikone  Let me know if this fixes it! And sorry about the trouble, not sure why I didn't run into this, but it's clearly an oversight on my part.

### DIFF
--- a/src/renderer/renderer.hpp
+++ b/src/renderer/renderer.hpp
@@ -93,17 +93,38 @@ public:
     {
     }
 
+    StateSaver(StateSaver&& other) noexcept
+      : mpRenderer(other.mpRenderer)
+      , mClipRect(other.mClipRect)
+      , mGlobalTranslation(other.mGlobalTranslation)
+      , mGlobalScale(other.mGlobalScale)
+      , mRenderTarget(other.mRenderTarget)
+    {
+      // Mark the moved-from object as such (see ~StateSaver)
+      other.mpRenderer = nullptr;
+    }
+
+    StateSaver& operator=(StateSaver&& other) noexcept {
+      if (&other != this) {
+        *this = std::move(other);
+      }
+
+      return *this;
+    }
+
     ~StateSaver() {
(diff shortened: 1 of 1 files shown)


pratikone:
Thanks, this definitely worked. I still have no idea what problem copy ellision caused here. Can you explain that to me ?

lethal-guitar:
> Can you explain that to me ?

@pratikone sure! So `Renderer::StateSaver` is a RAII class. It takes a snapshot of the current renderer state when constructed, and reapplies that copy of the state when destructed. The effect is that any changes to the renderer state that happen _after_ constructing a `StateSaver` will be undone when the `StateSaver` goes out of scope.

In the function `setupSimpleUpscaling`, I create a state saver first, make some changes to the renderer state, and then return the state saver from that function. The calling code stores the returned saver in a variable. Once that variable goes out of scope, the renderer state is supposed to be restored to what it was before calling `setupSimpleUpscaling`.

For this to work, returning the state saver from the function must *not* change the renderer state. This requires implementing a move constructor for the state saver, so that the saver instance that was created inside the function does nothing after it was moved into the return value.

I did this incorrectly at first by `default`-ing the move ctor, but that doesn't work (the default move ctor simply does a move of all members, which has no effect on primitive types).

But because of copy elision, I didn't notice my mistake: The compiler would apply RVO (return value optimization) and bypass the move ctor, resulting in the state saver (constructed inside the function) to stay alive until the variable holding the function's return value goes out of scope. Note that the compiler is allowed to skip any side-effects of move/copy ctors when doing copy elision.

So I made a mistake, but because of copy elision/RVO, everything was working correctly for me despite the mistake. This was only by chance though, different compilers could make slightly different optimization decisions and expose the bug.

PR #422, opened 2019-08-14
Create additional READMEs

lethal-guitar:


lethal-guitar:
@pratikone @s-martin I wrote some additional READMEs as a start for documenting the code base better. What do you think? Are these helpful?

pratikone:
These are definitely helpful.  You can add them as wiki pages (which can be easily edited/updated,) and link it in the main README as URL something like Understanding  Codebase / Contributing to the codebase

lethal-guitar:
Yeah I'll put it on the wiki too, and link from the top-level README. I like having the READMEs because they show up when you browse the code on GitHub, so I'll keep both.

PR #423, opened 2019-08-14
Fix widescreen mode for low resolution 16:10 screen.

lethal-guitar:
Doing some more testing, I noticed that the viewport calculation for widescreen mode didn't use the full available width of the screen on my laptop, which has a 1280x800 screen (16:10 aspect ratio). With that resolution, the horizontal scale factor works out to 3.33333..., which means we can fit 48 tiles into 1280 pixels (1 tile = 8 px, thus 8*3.333... -> 26.6666. 1280/26.6666 -> 48). But due to rounding too early, the calculation would work out to 47 tiles instead, leaving a gap of 13 pixels on each side of the screen.

### DIFF
--- a/src/renderer/upscaling_utils.cpp
+++ b/src/renderer/upscaling_utils.cpp
@@ -20,6 +20,8 @@
 #include "data/game_traits.hpp"
 #include "renderer/renderer.hpp"
 
+#include <algorithm>
+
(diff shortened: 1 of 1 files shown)


lethal-guitar:
@pratikone FYI

PR #441, opened 2019-10-12
Fix JSON merging

### LINKED ISSUE #414: Prevent older versions of RigelEngine from erasing parts of the user profile
*Background*: RigelEngine stores saved games, high scores and options in a "user profile" file, encoded as JSON (compressed as MsgPack). When loading the profile, the JSON is deserialized into a C++ data structure. When saving, this data structure is serialized into JSON and the profile file is overwritten.

*Problem:* A user runs version B of RigelEngine, which has several options. The user configures all options to their liking, and the profile file is stored including all these options. Now, the user runs an older version of the game, version A, on the same machine. This version isn't aware of the options stored in the file, so it ignores entries that it doesn't know about when deserializing. When saving the profile from version A, it overwrites the file and all the options are lost.

*Possible solution:* We cannot patch old versions of RigelEngine once they are out in the wild, so the only way to protect newer versions from having their options erased from the profile is to rename the profile file for the version that introduces options. We would then import saved games and high scores from the "old" user profile file, but store to a different filename.

Starting with the version that introduces a new file name, we can be a bit smarter and implement compatibility. One idea would be to store the raw JSON data from the profile file in memory, and merge it with the serialized profile when saving. That way, entries from newer versions would be preserved when an older version writes to the profile.

lethal-guitar:
Closes #414

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,14 +32,14 @@ matrix:
           packages:
             - sdl2
             - sdl2_mixer
-            - llvm
+            - llvm@8
 
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then
-      export CC="/usr/local/opt/llvm/bin/clang";
+      export CC="/usr/local/opt/llvm@8/bin/clang";
       export CXX="$CC++";
-      export CPPFLAGS="-I/usr/local/opt/llvm/include";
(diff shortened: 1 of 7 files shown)


lethal-guitar:
@pratikone FYI, if you're interested in looking at this

lethal-guitar:
@pratikone I'm going to merge this now, as there a few critical fixes here on this branch. But feel free to leave some comments later if you'd like!

Issue #410, opened 2019-08-06
Implement Gamepad support


lethal-guitar:
At the moment, RigelEngine only supports keyboard controls. This issue is about adding gamepad support using the [SDL Gamecontroller API](https://wiki.libsdl.org/CategoryGameController). Most of the relevant code is in [`GameRunner::World::handlePlayerInput`](https://github.com/lethal-guitar/RigelEngine/blob/8a53940b01e1f4ee236cc6439f95291245bb3195/src/game_runner.cpp#L375).

Once the game works with controller inputs, it would be great if the menus could also be navigated using a controller. For that, see:

* [DukeScriptRunner::handleEvent](https://github.com/lethal-guitar/RigelEngine/blob/8a53940b01e1f4ee236cc6439f95291245bb3195/src/ui/duke_script_runner.cpp#L158): Handle directional inputs, as well as "A" for enter and "B" for Escape
* [EpisodeEndScreen::handleEvent](https://github.com/lethal-guitar/RigelEngine/blob/0e579c52c277d92f6f64bd94c69022f75b06e4b6/src/ui/episode_end_sequence.cpp#L88): Any button press should be treated like a keypress, and advance to the next image
*[IntroDemoLoopMode::handleEvent](https://github.com/lethal-guitar/RigelEngine/blob/0e579c52c277d92f6f64bd94c69022f75b06e4b6/src/intro_demo_loop_mode.cpp#L92): Here, we'll want to handle any button press for the `if (mCurrentStage == 0)` case as if a key was pressed. For the other case, we'll want to handle the "B" button like Escape.

_Note_: The options menu is handled by Dear ImGui, and enabling Gamepad support for that is a separate issue, as it needs to be done differently.

pratikone:
Can it be tagged as good first issue given it doesn't require much knowledge of the codebase ?

lethal-guitar:
@pratikone yeah good point, I think it can! I'll also add a bit of info soon.

PR #466, opened 2019-12-26
GameController functionality for playing Duke and menus

pratikone:
This PR adds game controller functionality to the engine. It also adds game controller support for most of the menus (not all especially those related to typing).
The sticks and dpad are used for movement (only dpad during the menus). B and A for jump, X and Right shoulder for fire.
For the menus, dpad for the movement, A to select and B for exit.
It has been tested with xbox one controller on Windows. Needs testing on other OSes and other controllers like PS 4 controller and Nintendo Switch one.

### DIFF
--- a/src/anti_piracy_screen_mode.cpp
+++ b/src/anti_piracy_screen_mode.cpp
@@ -42,7 +42,7 @@ std::unique_ptr<GameMode> AntiPiracyScreenMode::updateAndRender(
 
   const auto anyButtonPressed = any_of(begin(events), end(events),
     [](const SDL_Event& event) {
-      return event.type == SDL_KEYDOWN;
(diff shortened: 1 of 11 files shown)


pratikone:
Solves #410 Game controller support

lethal-guitar:
Thanks again for this contribution, @pratikone! :+1: I finally had a chance to test it. It works very nicely with my Xbox 360 controller, but the analog stick on my PS4 controller is a bit finicky to use - sometimes you get up/down when trying to move horizontally, and vice versa. I suspect this could be improved by tweaking the dead zone. I'm happy to look into it once I have time again.

I also had a few thoughts on improving controller support further, I'll create some issues to keep track of them.

PR #468, opened 2019-12-28
Add User Profile conversion tool

### LINKED ISSUE #460: Make userprofile as human-readable json file
Make userprofile file as user readable json file. It doesn't contain anything confidential. Having a human-readable non-serialized file will aid in debugging efforts as well as any user modifications. 

constexpr auto PREF_PATH_ORG_NAME = "lethal-guitar";
constexpr auto PREF_PATH_APP_NAME = "Rigel Engine";
constexpr auto USER_PROFILE_FILENAME = "UserProfile_v2.rigel";

Also, why not save the userprofile file local to executable ?

lethal-guitar:
Adds a new standalone executable called `UserProfileTool`, which can convert between binary and JSON encoding. This makes it easier to inspect and modify the user profile with a text editor, without changing the way the game itself stores the profile.

@pratikone what do you think?

Closes #460

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,5 +115,6 @@ target_compile_options(entityx PRIVATE
 
 enable_testing()
 
+add_subdirectory(modding_tools)
 add_subdirectory(src)
 add_subdirectory(test)
--- a/modding_tools/CMakeLists.txt
+++ b/modding_tools/CMakeLists.txt
@@ -1,0 +1,5 @@
+add_executable(UserProfileTool user_profile_tool.cpp)
+target_link_libraries(UserProfileTool PRIVATE
+    SDL2::Main
+    rigel_core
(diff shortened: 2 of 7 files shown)


pratikone:
This seems to be a good idea though my only concern is yet another piece of tool to be maintained(but that's not a big problem). 
Given that the project is intended to be modding-friendly, it should be put under mod tools. Rest looks good

lethal-guitar:
Thanks for having a look, @pratikone!

> my only concern is yet another piece of tool to be maintained

Yeah, it's true, I agree. But I think it's a good trade-off, it helps to keep the main game itself simpler. And the tool is very small, so I hope that any maintenance efforts will be minimal.

> it should be put under mod tools

Sure, I'll rename the folder to `modding_tools`.

PR #481, opened 2020-01-02
Add GIF with gameplay to README

lethal-guitar:
@pratikone FYI


pratikone:
We should comeplete this

lethal-guitar:
I haven't merged this yet, because I'm not sure anymore how I feel about it. On one hand, the animation is cool, but on the other hand, it can also be quite distracting when you want to focus on reading the text. In an ideal world, there would be a pause button for these gifs, so everyone can decide if they want to see it or not.

I wanted to do some "user testing", show both versions of the README to a couple of people, and see which one is preferred. But didn't get around to doing that yet.

Issue #413, opened 2019-08-07
Implement "select game path" dialog on startup


lethal-guitar:
Currently, users have to either place the RigelEngine executable inside their Duke Nukem 2 installation, or pass the right path on the command line. Not everyone is comfortable using the latter, and it would be nice to make this easier in general.

The idea here is that if no game path is provided or the provided path doesn't exist, RigelEngine would show a little file selection dialog to let the user choose their Duke 2 installation.

Challenge with this: How to implement in a platform independent way without pulling in a huge dependency?

And a question: Should it use the in-game UI (Dear ImGui) or a native OS UI (before opening the OpenGL window)? This will influence which options are available.

Some options for native UI outside the game window would be:

* Provide native implementations for Windows and OS X only, using their respective APIs
* Evaluate and use https://github.com/mlabbe/nativefiledialog

Options for in-game UI:
* Roll our own based on `std::filesystem` and Dear ImGui
* Evaluate and use https://github.com/AirGuanZ/imgui-filebrowser

PR #495, opened 2020-03-11
Folder browser

pratikone:
This PR has the functionality to show a Imgui based folder browser for selecting the game path. This makes Rigel Engine more approachable for people who are not comfortable with terminal #413 .

This required allowing default arguments for gamepath at command line level so that they can be empty and then later implemented in folder browser. If cmd line game path is provided then, folder browser code is ignored. Added better and informative error handling for game path.      

Future scope  :  

- Move folder browser to its own file (right now it has its own event handling system and doesn't use existing imgui event handling in main loop in order to avoid complexity. Code duplication is preferred over boiler plate and complex code :) )

- Add GUI to choose between Download or Browse. This needs download shareware task to be completed and has be optionally enabled.

- More streamlined way of creating user profile . And saving game path in user profile so that it is not asked everytime.


lethal-guitar:
Good stuff, thanks for doing this! I should have a chance to take a closer look and try it out this weekend.

Having a dedicated event loop makes sense, I agree that it's better to have some duplication in this case in exchange for overall simplicity. The user profile creation is indeed a bit awkward at the moment, saving the game path in the profile should make it much nicer.

src/game_main.cpp
```diff
@@ -270,7 +270,7 @@ void handleCheckingAndSettingGamePath(SDL_Window* pWindow, std::string& gamePath
 }
 
 
-void gameMain(StartupOptions& options) {
+void gameMain(const StartupOptions& options) {
```
pratikone: I am missing the point here for keeping const here if we are anyways copying it to another variable and using it for the entirety of the game.

src/game_main.cpp
```diff
@@ -295,7 +293,8 @@ void gameMain(StartupOptions& options) {
 
   ui::imgui_integration::init(
     pWindow.get(), pGlContext, createOrGetPreferencesPath());
-  
+  auto imGuiGuard = defer([]() { ui::imgui_integration::shutdown(); });
```
pratikone: I agree this is much better. I wanted to do with scope guard too but I didn't think hard enough to implement it this elegantly.

src/game_main.cpp
lethal-guitar: This is `const` because that's what makes sense from the point of view of the code calling `gameMain`. In `main.cpp`, we create a `StartupOptions` object and pass it into `gameMain`. We don't need to know what `gameMain` does with that object, and we are not interested in seeing any changes made to that object reflected in `main.cpp`.

If I see a function that takes a non-const reference parameter, I'm assuming that this function is going to modify the object I pass in, in order for me to then do something else with the modified object. Often, this is also called an "out parameter" or "in/out parameter", as the parameter is used to pass information out of the function. It means I'm looking at a function that represents an intermediate step towards some final goal, and I would typically have other code using the object after this function call. An example would be something like `void removeEvenNumbers(std::vector<int>& vec)`.

This is not the case with `gameMain` - after the call to `gameMain(config);`, we don't care anymore about the `config` object. We are giving `config` to `gameMain` to provide it with information, but we don't need to know what it does with that information. It is purely an "in parameter", which gives input to the function, but we don't use it to retrieve output. Using a `const&` signals that to the reader.

I aim to lean towards a functional style, where most objects are immutable. Making a copy in order to modify it is very natural in the functional paradigm, and allows to keep the function signature immutable while still modifying what we are given. Semantically, this also makes sense to me in that the `StartupOptions` are a bit like a request to `gameMain`, which it may or may not honor. So `gameMain` may decide that it wants to change some aspects of the config, but this doesn't change the initial config request we provided.

There is another aspect, which is that a non-const reference prevents the usage of r-values. If the signature is `gameMain(StartupOptions&)`, we would not be able to do something like `gameMain(createConfig())`, because the return value of a function is an r-value. Note that Visual Studio was not standards compliant for many years, and did allow passing r-values to non-const references even though other compilers never allowed it. But this has changed in VS 2017 with the introduction of the `/permissive-` switch. There's more info in this [C++ Team Blog post](https://devblogs.microsoft.com/cppblog/permissive-switch/).

That was a lot of explanation 😅, does it make sense? FWIW, I also think this copy is going to go away once I'm done with making the changes I have planned. I have a pretty good idea of how to arrange the user profile creation in a way that fits well with the folder browser feature, so I think I could get that done soon.

lethal-guitar:
@pratikone thanks again for this contribution! As you've already seen, I'm making some changes. I'm not quite done yet, but there's not too much left.. Stay tuned 🙂

src/game_main.cpp
pratikone: Yeah, thanks for explaining it. I am definitely not aware of copy and use paradigm of functional programming. The r-value part also makes sense.

lethal-guitar:
@pratikone alright, it took a bit more time than I first thought, but it's quite solid now I think. Here's what I've added:

* I've refactored the user profile management code a little bit as preparation, that's the commits from a288ad5 to 3fe0858.
* The game path is now stored in the user profile (7539744)
* If an exception occurs after ImGui has been initialized, I show a little message box using ImGui instead of printing to the console (9ef0871)

I also made a few more changes to your code:

* Renamed `selectPathUsingImGuiFolderBrowser` to `runFolderBrowser`
* Renamed the variable `folder_browser` to `folderBrowser`, to follow the existing naming convention
* Added a `glClear` call to the beginning of the ImGui event loop, to make sure the background is not showing some random data and to make moving/resizing the file browser window work
* Renamed `handleCheckingAndSettingGamePath` to `setupForFirstLaunch`, and extended the logic to cover a few different cases. I've added a couple of comments, so hopefully it's fairly self-explanatory. I would imagine that the future shareware download code would also be called from that function.

Let me know what you think! I've done some testing and all three cases handled in `setupForFirstLaunch` are working fine as far as I can tell, but I guess it doesn't hurt to do some more testing.

lethal-guitar:
@pratikone originally, you had added `CMakeSettings.json` to the `.gitignore`. Do you remember where that file came from (which IDE/tool)? Is it covered by the `.vscode` entry?

Btw, you can also add local ignore entries to your git working copy, without committing a change to `.gitignore`, by adding entries to `.git/info/exclude`. Here's an [article](https://www.jerriepelser.com/blog/exlude-git-files-locally/) with more info on that git feature.

src/game_main.cpp
```diff
@@ -217,6 +256,138 @@ std::optional<FpsLimiter> createLimiter(const data::GameOptions& options) {
   }
 }
 
+
+std::string effectiveGamePath(
```
pratikone: As it chooses between option provided gamepath and userprofile provided game path, a comment would be better in explaining that.

src/game_main.cpp
```diff
@@ -217,6 +256,138 @@ std::optional<FpsLimiter> createLimiter(const data::GameOptions& options) {
...
+  const StartupOptions& options,
+  const UserProfile& profile
+) {
+  using namespace std::string_literals;
+
+  if (!options.mGamePath.empty()) {
+    return options.mGamePath;
+  }
```
pratikone: Should userprofile's game path be updated if a valid game path is provided via options ?

pratikone: It would also help future scenarios where a user has shareware path added in user profile and wants to add full version path instead.

pratikone:
Hey, thanks for doing the edits. These are some really good changes. It solves quite many of the profile and game path related issues.  Overall, looks good.

For the .gitignore, I will check the link to see how to keep it local. I have used Visual Studio and VS Code for coding Rigel Engine so it must have come from VS .

src/game_main.cpp
```diff
@@ -209,6 +211,43 @@ std::unique_ptr<GameMode> createInitialGameMode(
 }
 
 
+void showErrorMessage(SDL_Window* pWindow, const std::string& error) {
```
pratikone: is there a better place to move this ? like utility namespace ?

lethal-guitar: Yep, let's move it once we start using it in other places :+1:

src/game_main.cpp
lethal-guitar: Good question. I think there should definitely be a way for the user to switch to a different game path, switching from Shareware to registered version is a good example. But maybe this could also be done in the UI, like a button "Choose new game path" in the options menu?

Now that we have the game path browser in the UI, the command line game path looks like a feature for developers to me, where you might want to temporarily use a different path for testing, but use the one from the profile most of the time.

So for now, my suggestion would be that we don't update the profile, and I'll add an issue for the button to switch game path.

src/game_main.cpp
lethal-guitar: Is https://github.com/lethal-guitar/RigelEngine/pull/495/commits/b89674100d0d7e9ad9ac5b82e7ebbef5f1173e76 roughly what you had in mind?

src/game_main.cpp
lethal-guitar: I've already started working on choosing a different path from the options menu: https://github.com/lethal-guitar/RigelEngine/commits/game-path-switch

src/game_main.cpp
pratikone: indeed. thanks for adding it

PR #496, opened 2020-03-15
Fix OS X build on CI

lethal-guitar:


pratikone:
interesting update. how changing xcode version fixed it for MacOs ?

lethal-guitar:
It seems that Travis slightly changed the way they parse the configuration. Previously, having `osx_image` at the root level of the yaml file was working fine, but now it seems that it needs to be inside the `matrix` section for it to be picked up. So the same code that was building successfully earlier was now failing - it's a bit unfortunate, but it's something you have to deal with from time to time with these CI services.

As for why we need a specific version: On Mac OS, the C++ standard library is part of the operating system, which means you need a specific minimum OS version if you want to use modern C++ features. By default, Travis uses OS X version 10.13, but some C++ 17 features require at least version 10.14. With Travis, the OS X version is determined by the Xcode version, and when we specify `10.2` for Xcode, we are guaranteed to get OS X version 10.14 (see [MacOS version overview](https://docs.travis-ci.com/user/reference/osx/#macos-version)).

We were already using that version before, but due to the change in configuration parsing, Travis would fall back to the default of 10.13, making the build fail.

pratikone:
thanks for explaining so well.

PR #498, opened 2020-03-27
Allow switching game path from options menu

lethal-guitar:
This adds a button to the options menu, which allows choosing a different game path using the UI. Handy when a user wants to switch from a shareware install to a registered version install, for example.

The button is only shown in the main menu's options menu. When a valid new path is chosen, the game restarts itself to load resources from the new path, and stores the new path in the user profile.

### DIFF
--- a/src/data/image.cpp
+++ b/src/data/image.cpp
@@ -73,7 +73,7 @@ void Image::insertImage(
     throw invalid_argument("Source image doesn't fit");
   }
 
-  auto sourceIter = pixels.cbegin();
+  auto sourceIter = pixels.begin();
   for (size_t row=0; row<inferredHeight; ++row) {
     for (size_t col=0; col<sourceWidth; ++col) {
       const auto targetOffset = (x+col) + (y+row)*mWidth;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -307,6 +307,7 @@ target_link_libraries(rigel_core
     entityx
     Boost::boost
     dear_imgui
+    imgui-filebrowser
 
     PRIVATE
     std::filesystem
@@ -350,6 +351,5 @@ target_link_libraries(${main_exe_name} PRIVATE
     Boost::boost
     Boost::program_options
     Boost::disable_autolinking
-    imgui-filebrowser
 )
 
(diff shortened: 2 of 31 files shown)


lethal-guitar:
@pratikone check it out - this implements what we discussed [here](https://github.com/lethal-guitar/RigelEngine/pull/495#discussion_r395979515).

lethal-guitar:
Also, this makes me wonder if we should maybe show the entire options menu during the first launch, instead of just the folder browser? The options menu doesn't require game data to function. And this way, a user could already tweak some graphics options and have them stored in the profile, before selecting the game path.

Code-wise, it would unify the two places where we implement the folder browser into one place.

pratikone:
Well, either way would be fine. Showing or not showing. Though, it would be more user-friendly if user is able to launch and start a game without bothering much about options. We need file browser because the game cannot launch without resources. For options, we can keep good defaults and let user decide if they want to modify it or not. 
But, that doesn't mean that it is always the case. Like so many trade-offs, showing options on boot can work too if you think it will simplify handling of configurations and profile

src/data/image.cpp
```diff
@@ -73,7 +73,7 @@ void Image::insertImage(
     throw invalid_argument("Source image doesn't fit");
   }
 
-  auto sourceIter = pixels.cbegin();
+  auto sourceIter = pixels.begin();
```
pratikone: any particular reason of this change even when you are not modifying iter ?

lethal-guitar: Yes: If you have a `const vector` or other container, then `begin` and `end` will be equivalent to `cbegin`/`cend`. If you look at the [reference](https://en.cppreference.com/w/cpp/container/vector/begin) for `std::vector::begin`, for example, you can see that there are two overloads for `begin`:

```cpp
iterator begin() noexcept;
const_iterator begin() const noexcept;
const_iterator cbegin() const noexcept;
```

If `begin` is called on a `const` `vector`, the 2nd overload is chosen, and a `const_iterator` is returned. 

The case where `cbegin`/`cend` are useful is if you have a _non_ `const` container, but you want to get a `const_iterator` to it. When I originally wrote this code a couple years ago, this wasn't fully clear to me, so I used `cbegin`/`cend` everywhere. But that's not necessary, so I thought I would clean it up.

lethal-guitar:
> showing options on boot can work too if you think it will simplify handling of configurations and profile

I might give it a try at some point, and see how it looks and what impact it has on code complexity.

PR #526, opened 2020-05-24
Controller support improvements

### LINKED ISSUE #520: Game pad support enhancements
* When Duke is using his space ship, the deadzone for vertical analog stick movement should be the same as for horizontal.
* It should be possible to trigger teleporters, keycard readers etc. with pushing the analog stick up, in addition to pressing up on the d-pad
* It should be possible to navigate menus using the analog stick

lethal-guitar:
Closes #520

### DIFF
--- a/src/ui/duke_script_runner.cpp
+++ b/src/ui/duke_script_runner.cpp
@@ -18,6 +18,7 @@
 
 #include "base/container_utils.hpp"
 #include "base/match.hpp"
+#include "base/math_tools.hpp"
 #include "common/game_service_provider.hpp"
 #include "data/game_traits.hpp"
 #include "data/unit_conversions.hpp"
@@ -55,6 +56,8 @@ const auto SKILL_LEVEL_SLOT = 0;
 const auto GAME_SPEED_SLOT = 8;
 const auto INITIAL_SKILL_SELECTION = 1;
 const auto INITIAL_GAME_SPEED = 3;
+
+constexpr auto ANALOG_STICK_DEADZONE = 20'000;
 
 
 auto makeSpriteSheet(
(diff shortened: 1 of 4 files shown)


src/ui/duke_script_runner.cpp
```diff
@@ -56,6 +57,8 @@ const auto GAME_SPEED_SLOT = 8;
 const auto INITIAL_SKILL_SELECTION = 1;
 const auto INITIAL_GAME_SPEED = 3;
 
+constexpr auto ANALOG_STICK_DEADZONE = 20'000;
```
pratikone, line 60: I am seeing this number format for the first time, what does having a ‘ represent here ?

src/ui/duke_script_runner.cpp
```diff
@@ -233,33 +235,74 @@ void DukeScriptRunner::handleKeyboardEvent(const SDL_Event& event) {
 }
 
 void DukeScriptRunner::handleGameControllerEvent(const SDL_Event& event) {
-  if (event.type != SDL_CONTROLLERBUTTONDOWN) {
+  if (!hasMenuPages()) {
     return;
   }
 
-  // Menu interaction using dpad, button B and A
```
pratikone, line 240: This file can get refactored now. All non important functionality can be moved to a different file, that will shorten its size and reduce complexity due to obscurity issue

src/ui/duke_script_runner.cpp
lethal-guitar: It's a handy little feature from C++ 14. The `'` has no meaning for the compiler, it just separates the digits for better readability :slightly_smiling_face:

src/ui/duke_script_runner.cpp
lethal-guitar: It's a pretty big file, that's true. I have to admit that I don't see any immediately obvious way to split it up though. All of this code is required to implement the features of DukeScript. There's not much in terms of "non important" functionality here.

I can also see how this might be an odd/unexpected place to find the code for menu navigation, but that's because the original game uses DukeScript to implement its menus.

What I _could_ imagine is introducing a helper class to deal with the input handling. Something like a `MenuInputHandler`. And then other places in the code like `DukeScriptRunner` or elsewhere would talk to that class instead of handling SDL events directly. That would make it easier to change input handling without having to change a lot of different places in the code.

PR #531, opened 2020-05-27
Rigel Engine running in webassembly

pratikone:
This PR is a polished version of previous one where it enabled Rigel Engine to run on the web using webassembly. It uses emscripten library to cross compile to webassembly. The compilation targets are cleanly separated. webassembly related files are only compiled when running under emscripten. A new entry point is provided in emscripten_main.cpp when running under emscripten. More changes are newer variable WEBASSEMBLY_GAME_PATH for providing path of folder containing Duke Nukem II resources.  Currently, an html file `Rigel.html` is copied to build folder which loads the js and wasm. There are more options in Emscripten which can skip this part by directly creating an html file and using a template using `--shell-file`. It can be explored in the future
Image of rigel engine running in webassembly : https://imgur.com/gallery/vtbxjVo


3rd_party/imgui/CMakeLists.txt
```diff
@@ -14,10 +14,13 @@ target_include_directories(dear_imgui
...
 )
+# target_link_libraries(dear_imgui
+#     PRIVATE
+#     SDL2::Core
+#     glad
+# )
 target_link_libraries(dear_imgui
     PRIVATE
-    SDL2::Core
-    glad
```
lethal-guitar, line 20: Does it still work if you keep the `glad` dependency?

CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
 project(RigelEngine)
 
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -m32")
```
lethal-guitar, line 5: I'm unsure if this flag is actually necessary. When all the external dependencies are coming from Emscripten, and the other ones are built as part of the build, I would expect things to work without this. Could you give it a try?

CMakeLists.txt
```diff
@@ -13,21 +18,22 @@ set(main_exe_name RigelEngine)
 # Project settings
 ###############################################################################
 
-option(USE_GL_ES "Use OpenGL ES instead of regular OpenGL" OFF)
+option(USE_GL_ES "Use OpenGL ES instead of regular OpenGL" ON)
```
lethal-guitar, line 21: We should have some code to set the GL ES option to ON if building with Emscripten. That way, we can keep the default value at `OFF`.

CMakeLists.txt
```diff
@@ -13,21 +18,22 @@ set(main_exe_name RigelEngine)
...
 option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 
-
 # Dependencies
 ###############################################################################
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+# set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "/home/pratikone/git/emsdk/upstream/emscripten/cmake/Modules/")
```
lethal-guitar, line 28: Normally, the path should be set by the toolchain file, so this should not be necessary. There should also not be any harm in adding Rigel's own modules to the path (i.e. `cmake/Modules`), since we won't make use of them when in Emscripten mode. Could you check if it works when you keep the original module path instead of this one?

CMakeLists.txt
```diff
@@ -13,21 +18,22 @@ set(main_exe_name RigelEngine)
...
 
-set(Boost_USE_STATIC_LIBS ON)
-find_package(Boost 1.65 COMPONENTS program_options REQUIRED)
-find_package(SDL2 REQUIRED)
-find_package(SDL2_mixer REQUIRED)
+# set(Boost_USE_STATIC_LIBS ON)
```
lethal-guitar, line 30: All these `find_package` calls should go into an `if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")`

CMakeLists.txt
```diff
@@ -13,21 +18,22 @@ set(main_exe_name RigelEngine)
...
+# find_package(Boost 1.65 COMPONENTS program_options REQUIRED)
+# find_package(SDL2 REQUIRED)
+# find_package(SDL2_mixer REQUIRED)
 find_package(Filesystem REQUIRED)
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -std=c++1z -s WASM=1 -s USE_BOOST_HEADERS=1 -s USE_SDL=2 -s USE_SDL_MIXER=2 -s ALLOW_MEMORY_GROWTH=1 -s MIN_WEBGL_VERSION=1 -s MAX_WEBGL_VERSION=2 -s FULL_ES2=1 --preload-file /home/pratikone/Desktop/duke/")
```
lethal-guitar, line 36: I guess the preload files would be checked into the repo? Then it would be easy to make a relative path (using `${CMAKE_SOURCE_DIR}`)

modding_tools/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 add_executable(UserProfileTool user_profile_tool.cpp)
+# target_link_libraries(UserProfileTool PRIVATE
+#     SDL2::Main
+#     rigel_core
+# )
 target_link_libraries(UserProfileTool PRIVATE
-    SDL2::Main
```
lethal-guitar, line 3: It should be possible to define "dummy" targets for `SDL2` when building with Emscripten, that way, we would not have to change these. I'll look into it!

src/CMakeLists.txt
```diff
@@ -348,11 +374,13 @@ set(main_exe_sources
...
 add_executable(${main_exe_name} ${main_exe_sources})
+# target_link_libraries(${main_exe_name} PRIVATE
+#     SDL2::Main
+#     rigel_core
+#     Boost::boost
+#     Boost::program_options
+#     Boost::disable_autolinking
+# )
 target_link_libraries(${main_exe_name} PRIVATE
-    SDL2::Main
```
lethal-guitar, line 352: I have some ideas on how to split things up in a way that would allow us to define a different main target when building with Emscripten, I'll add more details some time later!

src/game_main.cpp
```diff
@@ -510,17 +535,7 @@ Game::Game(
   : mpWindow(pWindow)
   , mRenderer(pWindow)
   , mResources(effectiveGamePath(commandLineOptions, *pUserProfile))
-  , mIsShareWareVersion([this]() {
```
lethal-guitar, line 513: Do things work correctly without this change? I might be missing something, but if the game itself is working fine, then this check should also work fine in the Webassembly version.

src/game_main.hpp
```diff
@@ -18,7 +18,7 @@
 
 #include "common/command_line_options.hpp"
 
-
+#include "emscripten.h"
```
lethal-guitar, line 21: I think this can go into the `.cpp`, it shouldn't be needed in the header file. Also, we would need some conditional compilation here, so that it's only included when building with Emscripten

src/main.cpp
```diff
@@ -100,77 +100,77 @@ base::Vector parsePlayerPosition(const std::string& playerPosString) {
 
 
 int main(int argc, char** argv) {
-  showBanner();
+  // showBanner();
```
lethal-guitar, line 103: Do things still work if you leave in the call to `showBanner`? It's not really necessary, but it should not depend on `program_options`, so I'm curious if there was some other problem with it

src/main.cpp
```diff
@@ -100,77 +100,77 @@ base::Vector parsePlayerPosition(const std::string& playerPosString) {
...
 
     gameMain(config);
   }
-  catch (const po::error& err)
-  {
-    std::cerr << "ERROR: " << err.what() << "\n\n";
-    std::cerr << optionsDescription << '\n';
-    return -1;
-  }
+  // catch (const po::error& err)
```
lethal-guitar, line 168: I think we should have a separate `emscripten_main.cpp` file which is not using `program_options`, and then we can add either `main.cpp` or `emscripten_main.cpp` via CMake. That way, we don't need so much conditional compilation.

src/renderer/renderer.cpp
```diff
@@ -61,6 +60,7 @@ const auto SHADER_PREAMBLE = R"shd(
 const auto SHADER_PREAMBLE = R"shd(
 #version 150
 
+precision highp float;
```
lethal-guitar, line 63: This is the shader preamble for regular (non-ES) OpenGL, so it shouldn't need any changes, as it won't be used if `RIGEL_USE_GL_ES` is defined

src/renderer/renderer.cpp
```diff
@@ -188,7 +184,7 @@ vec4 applyWaterEffect(vec4 color) {
     }
   }
 
-  int adjustedIndex = (index & 0x3) | 0x8;
+  const int adjustedIndex = 0x8;//(index & 0x3) | 0x8;
```
lethal-guitar, line 187: I thought about this a bit, and I realized we can implement the same operation without using bit-wise operators like so:

```
int adjustedIndex = int(mod(float(index), 4) + 8);
```

That should be compatible with GL ES, and produce the same outcome.

CMakeLists.txt
lethal-guitar: Although that brings up the question whether it would be ok to check a copy of the Duke Shareware files into the repo, which I'm not sure about. This would be avoided by adding the automatic download feature, of course.

src/game_main.hpp
lethal-guitar: Another option would be to change things so that we could set the mainloop callback from `emscripten_main.cpp`, to keep `game_main.cpp` unaware of Emscripten. Since the former would only be built when building with Emscripten, it would then be ok to unconditionally include the Emscripten header.

The idea would be to add two more functions to `game_main.hpp`:

```cpp
std::unique_ptr<Game> createGame();
void runOneFrame(Game* game);
```

In the regular `main.cpp`, we would then continue calling `gameMain` as before, but in `emscripten_main.cpp`, we would do something like:

```cpp
void callback(void* pData) {
  rigel::runOneFrame(static_cast<rigel::Game*>(pData));
}

int main() {
  auto pGame = rigel::createGame();
  emscripten_set_main_loop_arg(callback, pGame.get(), 0, true);
  return 0;
}
```

This way, the different code paths are nicely separated. And we could even add some more code later to allow passing some command line options via URL parameters or something, if we wanted to.

CMakeLists.txt
pratikone: Yeah, it is not. Though I would like to keep it in some form because wasm64 doesn't exist AFAIK, for now. So for emscripten, I would like to make sure it builds only 32-bit

CMakeLists.txt
pratikone: there exists a __EMSCRIPTEN__ guard which I am planning to use

CMakeLists.txt
pratikone: https://emscripten.org/docs/compiling/Building-Projects.html#detecting-emscripten-in-preprocessor

src/renderer/renderer.cpp
pratikone: awesome

CMakeLists.txt
pratikone: Yup. we are in the similar situation as OpenLara project. They have shipped some form of files when porting Tomb Raider to webassembly http://xproger.info/projects/OpenLara/

CMakeLists.txt
lethal-guitar: For this, we can do it in CMake:

```
if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
  set(USE_GL_ES ON)
endif()
```

CMakeLists.txt
lethal-guitar: That should be the responsibility of the Emscripten toolchain file, however. The general philosophy of CMake is that Toolchain files are used to define flags specific to the target system. Project files (`CMakeListst.txt`) should ideally not contain any compiler flags at all. It's hard to achieve in practice (unless you would write and maintain your own toolchain files), but I would still like to follow it where possible.

The idea is that if wasm64 would be introduced one day, we would only have to update Emscripten, but not change any code in our `CMakeLists`. 

Take a look at [this line](https://github.com/emscripten-core/emscripten/blob/efe15600fc2d9210f4a151080f3eab76eec7499c/cmake/Modules/Platform/Emscripten.cmake#L26): The Emscripten toolchain file tells CMake that the platform is a 32-bit one.

CMakeLists.txt
lethal-guitar: Didn't know about that project! It's so cool to see how many of these reverse engineering projects there are nowadays :slightly_smiling_face: 

There's also the option to write a script `setup_wasm_build.sh`, which could take the path to the game files as an argument, run CMake with the right options, and then copy the game files into the build directory. That way, we can easily produce wasm builds with the game files included, but we don't have to check them into the repo.

CMakeLists.txt
lethal-guitar: > there is an equivalent variable for Cmake too

Yes, I'm using it in the code snippet in my comment above (checking `CMAKE_SYSTEM_NAME`). Or are you referring to a different variable?

src/renderer/renderer.cpp
lethal-guitar: I've made that change in #533

lethal-guitar:
@pratikone I've been thinking back and forth on how to do the main loop in a nice way, had various ideas, but in the end, we should probably do the simplest thing possible and accept a bit of code duplication. So my current idea is as follows:

First, introduce a separate `emscripten_main.cpp`, as I already mentioned in the comments:

```
# Main executable
if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
    add_executable(${main_exe_name} emscripten_main.cpp)
    target_link_libraries(${main_exe_name} PRIVATE
        rigel_core
    )
else()
    add_executable(${main_exe_name} main.cpp)
    target_link_libraries(${main_exe_name} PRIVATE
        SDL2::Main
        rigel_core
        Boost::boost
        Boost::program_options
        Boost::disable_autolinking
    )
endif()
```

To do that, all the files listed in `main_exe_sources` need to be moved to `core_sources` instead. But there's no reason why we shouldn't do that at this point.

Then, a bit of refactoring:

* We need to introduce `Game::runOneFrame`, which would do everything that is currently in the body of the for loop inside of `Game::mainLoop`. 
* Remove `Game::mainLoop`, and instead do the `for` in `initAndRunGame`.
* Merge the remaining code from `Game::run` into the constructor of `Game`.

With that out of the way, we can then add the `main` function in `emscripten_main.cpp`:

```cpp
void runOneFrameWrapper(void* pData) {
  static_cast<Game*>(pData)->runOneFrame();
}

int main() {
  sdl_utils::check(SDL_Init(
    SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_GAMECONTROLLER));
  auto sdlGuard = defer([]() { SDL_Quit(); });

  sdl_utils::check(SDL_GL_LoadLibrary(nullptr));
  setGLAttributes();

  auto userProfile = loadOrCreateUserProfile();
  auto pWindow = createWindow(userProfile.mOptions);
  SDL_GLContext pGlContext =
    sdl_utils::check(SDL_GL_CreateContext(pWindow.get()));
  auto glGuard = defer([pGlContext]() { SDL_GL_DeleteContext(pGlContext); });

  renderer::loadGlFunctions();

  SDL_DisableScreenSaver();
  SDL_ShowCursor(SDL_DISABLE);

  ui::imgui_integration::init(
    pWindow.get(), pGlContext, createOrGetPreferencesPath());
  auto imGuiGuard = defer([]() { ui::imgui_integration::shutdown(); });

  CommandLineOptions options;
  options.mGamePath = /* add the hard-coded path into the pre-loaded data here */;
  Game game{options, &userProfile, pWindow.get()};

  emscripten_set_main_loop_arg(runOneFrameWrapper, &game, 0, true);
}
```

Let me know what you think, is this enough information to get started?

CMakeLists.txt
pratikone: no the same one

CMakeLists.txt
pratikone: yeah, open source clone of Tomb Raider has existed since 2013 atleast. They are doing pretty good covering TR 1-5.

pratikone:
> @pratikone I've been thinking back and forth on how to do the main loop in a nice way, had various ideas, but in the end, it all seemed too complicated. I think we should probably do the simplest thing possible and accept a bit of code duplication initially. We can then see if we can reduce it later. So my current idea is as follows:
> 
> First, introduce a separate `emscripten_main.cpp`, as I already mentioned in the comments:
> 
> ```cmake
> # Main executable
> if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
>     add_executable(${main_exe_name} emscripten_main.cpp)
>     target_link_libraries(${main_exe_name} PRIVATE
>         rigel_core
>     )
> else()
>     add_executable(${main_exe_name} main.cpp)
>     target_link_libraries(${main_exe_name} PRIVATE
>         SDL2::Main
>         rigel_core
>         Boost::boost
>         Boost::program_options
>         Boost::disable_autolinking
>     )
> endif()
> ```
> 
> To do that, all the files listed in `main_exe_sources` need to be moved to `core_sources` instead. But there's no reason why we shouldn't do that at this point.
> 
> Then, a bit of refactoring:
> 
>     * We need to introduce `Game::runOneFrame`, which would do everything that is currently in the body of the for loop inside of `Game::mainLoop`.
> 
>     * Remove `Game::mainLoop`, and instead do the `for` in `initAndRunGame`.
> 
>     * Merge the remaining code from `Game::run` into the constructor of `Game`.
> 
> 
> With that out of the way, we can then add the `main` function in `emscripten_main.cpp`:
> 
> ```c++
> void runOneFrameWrapper(void* pData) {
>   static_cast<Game*>(pData)->runOneFrame();
> }
> 
> int main() {
>   sdl_utils::check(SDL_Init(
>     SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_GAMECONTROLLER));
>   auto sdlGuard = defer([]() { SDL_Quit(); });
> 
>   sdl_utils::check(SDL_GL_LoadLibrary(nullptr));
>   setGLAttributes();
> 
>   auto userProfile = loadOrCreateUserProfile();
>   auto pWindow = createWindow(userProfile.mOptions);
>   SDL_GLContext pGlContext =
>     sdl_utils::check(SDL_GL_CreateContext(pWindow.get()));
>   auto glGuard = defer([pGlContext]() { SDL_GL_DeleteContext(pGlContext); });
> 
>   renderer::loadGlFunctions();
> 
>   SDL_DisableScreenSaver();
>   SDL_ShowCursor(SDL_DISABLE);
> 
>   ui::imgui_integration::init(
>     pWindow.get(), pGlContext, createOrGetPreferencesPath());
>   auto imGuiGuard = defer([]() { ui::imgui_integration::shutdown(); });
> 
>   CommandLineOptions options;
>   options.mGamePath = /* add the hard-coded path into the pre-loaded data here */;
>   Game game{options, &userProfile, pWindow.get()};
> 
>   emscripten_set_main_loop_arg(runOneFrameWrapper, &game, 0, true);
> }
> ```
> 
> Let me know what you think, is this enough information to get started?

3rd_party/imgui/CMakeLists.txt
pratikone: it does. just tested. adding it back

src/game_main.cpp
```diff
@@ -408,7 +409,18 @@ void initAndRunGame(
...
+    std::vector<SDL_Event> eventQueue;
+    game.mLastTime = high_resolution_clock::now();
+    for(;;){
+      bool breakOut = false;
+      game.runOneFrame(eventQueue, breakOut);
+      if (breakOut){
+        break;
+      }
+    }
+    return Game::RunResult::GameEnded; //TODO : figure out restart case
```
lethal-guitar, line 423: To handle the restart case, my suggestion would be to:

* Add `RunResult::StillRunning` to the `RunResult` enum
* Make `Game::runOneFrame` have a `RunResult` return value

Then, this code here would become:

```
for (;;) {
  const auto result = game.runOneFrame(eventQueue);
  if (result != RunResult::Still running) {
    return result;
  }
}
```

modding_tools/CMakeLists.txt
pratikone: can you elaborate on that ? What do you mean by dummy targets ? My cmake-fu is still not strong enough :)

src/game_main.cpp
pratikone: yup, can do that. I will make this change meanwhile you can review rest of the parts of the PR

modding_tools/CMakeLists.txt
lethal-guitar: Sure! So every time you say `add_library` or `add_executable` in CMake, you define a "target". Most of the time, a target results in code being compiled and a executable or library being produced. For example when we say `add_library(rigel_core`, it tells CMake to generate build files (Make files, VS solution etc.) that will compile all the sources listed for `rigel_core` and then link them into a static library.

Targets can depend on other targets using `target_link_libraries`. We use this to make the `RigelEngine` executable depend on the `rigel_core` library.

But targets are a more general concept, they don't always have to involve compiling code. A target can also be used to specify certain compiler flags, include paths, linker flags etc. Any target that depends on another target will then inherit all these compiler flags etc.

This is used in CMake for external dependencies. For example, we use `target_link_libraries` to specify that the target `rigel_core` should depend on the target `SDL2::Core`. The latter is a target  as well, just like `rigel_core`, but it does not cause any code to be compiled. Instead, it only defines the necessary include paths and compiler flags that need to be used in order to use the external SDL2 library. By saying that `rigel_core` depends on `SDL2::Core`, CMake knows that it needs to add all the compiler flags defined by `SDL2::Core` to any compiler command invocation that happens when building the files that are part of `rigel_core`.

Now where does the target `SDL2::Core` come from? It's defined by the find script, which we invoke when we use the `find_package(SDL2)` command. You can see how it's defining the target here: https://github.com/lethal-guitar/RigelEngine/blob/fbe161346b90ca8a43b74845d27a256a65fac0e7/cmake/Modules/FindSDL2.cmake#L354

Now my idea was that for the Emscripten case, we can define a target named `SDL2::Core` ourselves, and that target would specify that the `USE_SDL=2` flag should be passed to the compiler. Then we don't need to change any of the CMake code that uses the `SDL2::Core` target.

It would probably look something like this (not tested):

```cmake
add_library(SDL2::Core UNKNOWN IMPORTED)
set_target_properties(SDL2::Core PROPERTIES
  INTERFACE_COMPILE_OPTIONS "-s USE_SDL=2")
```
We could then do the same for SDL2_mixer and Boost.

Let me know if that helps!

AUTHORS.md
```diff
@@ -8,7 +8,7 @@ Nikolai Wuttke ([@lethal-guitar](https://github.com/lethal-guitar))
 
 * Alain Martin ([@McMartin](https://github.com/McMartin)): Help with OS X port, clang-tidy, compiler upgrades and lots of other things
 * Evo Stamatov ([@avioli](https://github.com/avioli)): README improvements
-* Pratik Anand ([@pratikone](https://github.com/pratikone)): Widescreen support, README improvements
+* Pratik Anand ([@pratikone](https://github.com/pratikone)): Webassembly, widescreen support, file browser, gamepad, README improvements
```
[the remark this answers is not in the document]
pratikone: yeah, feel free to modify even my contributions to better representation

lethal-guitar:
@pratikone I've made a new branch with all the necessary main loop refactoring done: https://github.com/lethal-guitar/RigelEngine/pull/561

I would suggest that you make a new PR based on that new branch, and re-add the Webassembly changes as new commits on top of that. The only required changes should be adding `emscripten_main.cpp` and the CMake changes, everything else should already be in place now. That would make it much easier for me to merge your changes 🙂

src/game_main.ipp
```diff
@@ -69,15 +108,16 @@ public:
   Game(const Game&) = delete;
   Game& operator=(const Game&) = delete;
 
+  std::chrono::high_resolution_clock::time_point mLastTime;
```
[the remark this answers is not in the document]
lethal-guitar: I did that already in https://github.com/lethal-guitar/RigelEngine/pull/561

src/emscripten_main.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <emscripten.h>
+
+using namespace rigel;
+using namespace rigel::misc;
+
+// Emscripten only takes a single argument. Therefore, all the needed arguments are combined in this struct and sent
+struct em_args
+{
+  Game &game;
+  std::vector<SDL_Event> &eventQueue;
```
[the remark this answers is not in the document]
lethal-guitar: I turned it into a member in https://github.com/lethal-guitar/RigelEngine/pull/561.

src/game_main.cpp
lethal-guitar: I've done that now in https://github.com/lethal-guitar/RigelEngine/pull/561.

src/emscripten_main.cpp
```diff
@@ -0,0 +1,95 @@
...
+RIGEL_DISABLE_WARNINGS
+#include <boost/algorithm/string/classification.hpp>
+#include <boost/algorithm/string/split.hpp>
+RIGEL_RESTORE_WARNINGS
+
+#include <iostream>
+#include <stdexcept>
+#include <string>
+#include "game_main.hpp"
+#include "game_main.ipp"
```
lethal-guitar, line 33: `.ipp` files are normally meant to be an implementation detail. Instead of including it here, it should be included at the end of `game_main.hpp`. But in https://github.com/lethal-guitar/RigelEngine/pull/561, I've added some functions that make it possible to create the game and call `runOneFrame` without needing to include the `.ipp` at all (it's the last commit).

src/game_main.cpp
```diff
@@ -50,36 +48,11 @@ namespace rigel {
...
 using namespace engine;
 using namespace sdl_utils;
+using namespace std::chrono;
 
 using RenderTargetBinder = renderer::RenderTargetTexture::Binder;
 
-namespace {
-
-template <typename Callback>
-class CallOnDestruction {
```
lethal-guitar, line 59: In https://github.com/lethal-guitar/RigelEngine/pull/561, I've moved this to a dedicated new file

src/game_main.cpp
```diff
@@ -50,36 +48,11 @@ namespace rigel {
...
-  Callback mCallback;
-};
-
-
-template <typename Callback>
-[[nodiscard]] auto defer(Callback&& callback) {
-  return CallOnDestruction{std::forward<Callback>(callback)};
-}
-
+namespace misc{
```
lethal-guitar, line 55: I decided to introduce a new `platform` namespace instead of `misc`, see https://github.com/lethal-guitar/RigelEngine/pull/561

AUTHORS.md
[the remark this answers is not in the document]
lethal-guitar: Updated in https://github.com/lethal-guitar/RigelEngine/commit/65280a395cf9119788149d282a48deb3bba14bc1

pratikone:
@lethal-guitar sounds good. I will port the changes today and tomorrow to the new branch and send a new PR.

src/emscripten_main.cpp
pratikone: ok, I always wondered the reasoning behind ipp files. Are they part of PIMPLE kind of pattern ?

lethal-guitar: You could use them for the pimpl-pattern, but it's a more general convention. The general use case is when you have inline functions or templates. If you have "normal" functions and classes, you have declarations in the header file, and implementation in the `.cpp` file. The interface is very readable and clean this way, because the header only contains the most important details.

But for templates and inline functions, you need to put your code into the header file. That can hurt readability. By moving the implementations into a separate `.ipp` file, and then including that file at the end of the header file, you can keep the interface nice and readable, while still making all the code visible to the compiler thanks to including the `.ipp` in the header.

That's the typical convention. I sometimes use `.ipp` for slightly different purposes, though, and this is also the case for `game_main.ipp`. In this case, I wanted to keep the declaration of the `Game` class separate from the implementation in the `.cpp` file, but I also didn't want to expose the `Game` class to the code that includes `game_main.hpp`, mainly for better compilation times. If `game_main.hpp` would include `game_main.ipp`, then `main.cpp` would transitively include all the headers included by `game_main.ipp`, and that would increase compile time for `main.cpp`.

The functions I introduced for creating a `Game` and calling `runOneFrame` use a forward declaration of the `Game` class, and thus don't need the declaration of `Game` to be available. It's a very similar technique to the pimpl-pattern.

PR #566, opened 2020-07-06
Webassembly changes for running Rigel Engine in a web browser

pratikone:
This PR is a refactored and reduced version of #531 where it enabled Rigel Engine to run on the web using webassembly. It uses emscripten library to cross compile to webassembly. The compilation targets are cleanly separated. webassembly related files are only compiled when running under emscripten. A new entry point is provided in emscripten_main.cpp when running under emscripten. More changes are newer variable WEBASSEMBLY_GAME_PATH for providing path of folder containing Duke Nukem II resources. Currently, an html file Rigel.html is copied to build folder which loads the js and wasm. There are more options in Emscripten which can skip this part by directly creating an html file and using a template using --shell-file. It can be explored in the future
Image of rigel engine running in webassembly : https://imgur.com/gallery/vtbxjVo

### DIFF
--- a/emscripten/Rigel.html
+++ b/emscripten/Rigel.html
@@ -1,0 +1,26 @@
+<!doctype html>
+<html lang="en-us">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <style>
+      html, body, canvas {
+        width: 100%;
+        height: 100%;
+        margin: 0;
+        padding: 0;
+        background-color: black;
+      }
+    </style>
+  </head>
+  <body>
+    <canvas id="canvas" oncontextmenu="event.preventDefault()"></canvas>
+    <script>
+      Module = {
+        canvas: document.getElementById('canvas')
+      }
+    </script>
+    <script async type="text/javascript" src="RigelEngine.js"></script>
+  </body>
+</html>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,10 +22,51 @@ option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
 
-set(Boost_USE_STATIC_LIBS ON)
-find_package(Boost 1.65 COMPONENTS program_options REQUIRED)
-find_package(SDL2 REQUIRED)
-find_package(SDL2_mixer REQUIRED)
+
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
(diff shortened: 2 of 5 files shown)


lethal-guitar:
Nice one! :+1:

lethal-guitar:
@pratikone the changes for the main `CMakeLists.txt` were missing on this branch. But that's no problem, I added them back and also did a couple of cleanups that I had mentioned on the other PR - let me know what you think about these changes. I'll squash all those commits into the first one before merging.

I also managed to get it running locally! There are some issues, however:

* I have no sound in Chrome on Mac, or in Firefox on Windows. Sound works in Firefox on Mac, but it's very distored and the music is too fast. In Chrome, there's also a warning in the console about WebAudio requiring a user input to be activated.
* The top of the image is cut off (see screenshot)
* Loading times are quite long, like when going from the intro movie to the menu etc. This is odd, as all the data should already be in memory. But I guess the format conversion that's pretty much instant in native code is much slower in Wasm somehow? Needs more investigation..

grafik

I'm happy to merge the branch despite these issues, all of this can be improved over time. But maybe we should mention in the README that it might not work perfectly everywhere yet?

pratikone:
Yeah, there are definitely few issues to be resolved which require more emscripten understanding. As for cutoff, that is due to screen refresh. If you fullscreen and back, it would be normal. In windows firefox, emscripten tries to move you into fullscreen as soon as the game receives focus. So that fixes it. For the sound, it was almost normal on windows. I encountered audio issues only on Linux Firefox.    
Thanks for doing the missing changes. I was planning to do them today once I realized yesterday that they were remaining. It sounds good. I suggest opening issues for wasm parts, I can infact start asking in emscripten community on reddit and elsewhere, how they are fixing it. Having a hosted demo will help in that case.

lethal-guitar:
>  As for cutoff, that is due to screen refresh. It needs a screen refresh to readjust the size. If you fullscreen and back, it would be normal. In Firefox, emscripten tries to move you into fullscreen automatically as soon as the game receives focus. So that always fixed it for me. Even resizing the browser should fix it. 

Interesting! I tried that, but it doesn't fix it for me unfortunately 🙁

lethal-guitar:
Alright, we have a bunch of issues now all collected in the [Webassembly milestone](https://github.com/lethal-guitar/RigelEngine/milestone/16). I also pushed one more commit with some tweaks to the README. If that looks good to you as well, I'll merge :tada:

emscripten/Rigel.html
```diff
@@ -0,0 +1,26 @@
...
+  <body>
+    <canvas id="canvas" oncontextmenu="event.preventDefault()"></canvas>
+    <script>
+      Module = {
+        canvas: document.getElementById('canvas')
+      }
+    </script>
+    <script async type="text/javascript" src="RigelEngine.js"></script>
+  </body>
+</html>
```
pratikone, line 26: i don't know why it doesn't have EOF but can you fix that ?

pratikone:
Looks all good

emscripten/Rigel.html
lethal-guitar: Sure thing, done! 👍

PR #574, opened 2020-07-12
Fix Webassembly audio

### LINKED ISSUE #568: Webassembly: Sound issues in Firefox
Issues with sound I observed on the machines I've tested on so far:

* On Windows, I have no sound at all.
* On Mac, sound plays but the music is way too fast and it sounds distorted overall.

The latter is possibly a frequency mismatch. We might need to be more robust about how we deal with the audio parameters. As mentioned in the docs for [`Mix_QuerySpec`](https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer_15.html#SEC15):

> Get the actual audio format in use by the opened audio device. This may or may not match the parameters you passed to Mix_OpenAudio.

Which means it's possible that the actual sample rate is different from what we request when calling `Mix_OpenAudio`, which would make the resampling and AdLib emulation output wrong data. We should call `Mix_QuerySpec` after opening, and use the actual sample rate and format reported by the library for further processing.

lethal-guitar:
Fixes #568 

Todo:

* [x] Return early from ctor if `Mix_OpenAudio` fails

### DIFF
--- a/src/data/sound_ids.hpp
+++ b/src/data/sound_ids.hpp
@@ -65,27 +65,14 @@ enum class SoundId {
   IntroDukeSpeaks2 = 40
 };
 
-
-namespace detail {
-
-const auto FirstSoundId = SoundId::DukeNormalShot;
-const auto LastSoundId = SoundId::IntroDukeSpeaks2;
-
-inline SoundId nextSoundId(const SoundId id) {
-  const auto nextIndex = static_cast<int>(id) + 1;
-  return static_cast<SoundId>(nextIndex);
-}
-
-}
+constexpr auto NUM_SOUND_IDS = 41;
 
 
 template<typename Callable>
 void forEachSoundId(Callable callback) {
-  using namespace data::detail;
-  for (auto id = FirstSoundId; id != LastSoundId; id = nextSoundId(id)) {
-    callback(id);
+  for (auto i = 0; i < NUM_SOUND_IDS; ++i) {
+    callback(static_cast<SoundId>(i));
   }
-  callback(LastSoundId);
 }
 
 }
--- a/src/engine/sound_system.cpp
+++ b/src/engine/sound_system.cpp
@@ -93,6 +93,48 @@ void appendRampToZero(data::AudioBuffer& buffer) {
 }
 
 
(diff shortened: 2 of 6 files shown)


pratikone:
Ok, so what thing was missing in the audio ? I tried skimming through the fix but didn't get the idea of problem it was fixing. Is it the lack of buffer ?

lethal-guitar:
@pratikone the audio code didn't output the right format in all cases. With audio, you have various different ways of representing an audio signal as binary data. It's similar to pictures, where different ways of encoding colors exist: RGB, CMYK, HSV, etc., and some parameters: How many bits per pixel (8, 16, 24...)? What's the order of components (RGB, BGR, ..)? Imagine you have a bitmap with the colors stored in order B, G, R for each pixel. If I were to take the memory of that picture and give it to some other code that assumes the order to be R,G,B instead, I would get the wrong colors on screen.

This is basically what happened here: My code was assuming that audio is always stored as a series of signed 16-bit integers, but WebAudio expects the values to be floats. The conversion was missing, and so the browser was interpreting my 16-bit integers as floating point numbers, resulting in garbage audio data being processed. Interestingly, the music was still somewhat recognizable, but sounded distorted and too fast. This makes sense, since a float is 32 bits, so two consecutive 16-bit sample values would be interpreted as a single value instead, resulting in doubling the playback speed.

This was on one system, anyway - on my other machine, I had no sound at all, so maybe the browser detected the problematic data and muted the audio there - just a guess though.

The fix is to use [`SDL_ConvertAudio`](https://wiki.libsdl.org/SDL_ConvertAudio) to convert the data produced by RigelEngine from 16-bit integers (Rigel always uses that format internally) to the format expected by the OS/platform before handing it over to SDL. The code was always slightly incorrect before, but because most platforms are happy to take audio data as 16-bit integers, this problem never manifested until we ran it on a browser (it's a very common format, it's what most WAV files use, CD-Audio, etc.)

lethal-guitar:
I guess this is a good opportunity to add some comments to this code.. I'll do that before merging :slightly_smiling_face:

lethal-guitar:
@pratikone I've split up some changes from the commit with the fix into a separate commit, and added a bunch of comments. Hope that makes it a bit clearer what's going on.

PR #579, opened 2020-07-24
CMake cleanups

### LINKED ISSUE #390: Overhaul/modernize CMake files
* Instead of setting warnings globally and then disabling them for 3rd party code, define a target which configures warnings, and make only our own code depend on that target
* For 3rd party libs where we have our own CMake file, define targets in a way that it's sufficient to depend on those targets (i.e., it should not be necessary to separately specify an include dir or similar)

lethal-guitar:
Closes #390

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
 
 project(RigelEngine)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,28 +110,10 @@ else()
 endif()
 
 
-# 3rd party build targets
+# Build targets
 ###############################################################################
 
-set(ENTITYX_BUILD_SHARED false CACHE INTERNAL "")
-set(ENTITYX_BUILD_TESTING false CACHE INTERNAL "")
-set(ENTITYX_DT_TYPE double CACHE INTERNAL "")
-set(ENTITYX_RUN_BENCHMARKS false CACHE INTERNAL "")
-set(ENTITYX_MAX_COMPONENTS 64 CACHE INTERNAL "")
-
-add_subdirectory(3rd_party/dbopl)
-add_subdirectory(3rd_party/entityx)
-add_subdirectory(3rd_party/glad)
-add_subdirectory(3rd_party/speex_resampler)
-add_subdirectory(3rd_party/imgui)
-add_subdirectory(3rd_party/imgui-filebrowser)
-
-rigel_disable_warnings(entityx)
-
-
-# RigelEngine build targets
-###############################################################################
-
+add_subdirectory(3rd_party)
 add_subdirectory(src)
 
 if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,56 +66,7 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 
-if(MSVC)
-    add_compile_options(
-        /W4
-        /wd4100 # Unused parameter
-        /wd4503 # Decorated name length exceeded
-        /wd4800 # Forcing value to bool
-    )
-    add_definitions(-D_SCL_SECURE_NO_WARNINGS)
-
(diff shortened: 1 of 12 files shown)


lethal-guitar:
@McMartin just in case you are bored and feel like reviewing some CMake, let me know if you notice anything that could be improved further 🙂

McMartin:
@lethal-guitar I have a look now 👀

src/CMakeLists.txt
```diff
@@ -359,6 +359,7 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
       rigel_core
     )
     target_link_options(${main_exe_name} PRIVATE
```
McMartin: `target_link_options` was introduced in CMake 3.13. You should update the version in the call to `cmake_minimum_required` in the top-level `CMakeLists.txt` file.

CMakeLists.txt
```diff
@@ -34,7 +34,10 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
         message(STATUS "Bundling game data from path: ${WEBASSEMBLY_GAME_PATH}")
     endif()
 
-    add_compile_options(-O3)
+    if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
+        # Emscripten's CMake toolchain defaults to O2, but we want O3
+        add_compile_options(-O3)
+    endif()
```
McMartin, line 39:
```suggestion
    # Emscripten's CMake toolchain defaults to O2, but we want O3
    add_compile_options($<$<NOT:$<CONFIG:Debug>>:-O3>)
```
So it also works with multi-configuration generators, and the `MinSizeRel` and `RelWithDebSymbols` configurations.

cmake/rigel.cmake
```diff
@@ -0,0 +1,28 @@
...
+# set the appropriate compiler and linker flags for Emscripten instead of
+# referring to external libraries.
+function(rigel_define_wasm_targets_for_dependencies)
+    add_library(SDL2::Core INTERFACE IMPORTED)
+    set_target_properties(SDL2::Core PROPERTIES
+        INTERFACE_COMPILE_OPTIONS
+        "SHELL:-s USE_SDL=2"
+        INTERFACE_LINK_OPTIONS
+        "SHELL:-s USE_SDL=2"
+    )
```
McMartin, line 13:
```suggestion
    target_compile_options(SDL2::Core INTERFACE "SHELL:-s USE_SDL=2")
    target_link_options(SDL2::Core INTERFACE "SHELL:-s USE_SDL=2")
```
and similarly for `SDL2::Mixer` and `Boost::boost`.

cmake/rigel.cmake
```diff
@@ -26,3 +26,22 @@ function(rigel_define_wasm_targets_for_dependencies)
         "SHELL:-s USE_BOOST_HEADERS=1"
     )
 endfunction()
+
+
+function(rigel_disable_warnings target)
+    if (MSVC)
+        target_compile_options(${target} PRIVATE
+            /w
```
McMartin, line 88: MSVC supports both `/` and `-` for flags, so you could use `-w` with all compilers.

src/CMakeLists.txt
lethal-guitar: Good catch, thanks!

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 ###############################################################################
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+include("${CMAKE_SOURCE_DIR}/cmake/rigel.cmake/")
```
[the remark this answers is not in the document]
lethal-guitar: I'm surprised it was working like this 🤔

CMakeLists.txt
lethal-guitar: Nice! 👍

cmake/rigel.cmake
lethal-guitar: Yeah that looks much more readable, thanks!

cmake/rigel.cmake
lethal-guitar: Sounds good

lethal-guitar:
@McMartin thanks for the review, really nice suggestions!

PR #582, opened 2020-07-25
More Wasm improvements

### LINKED ISSUE #569: Webassembly: Screen is cut off
On all the systems I tested, the top of the image is cut off in the browser. It seems that resizing the window or switching to fullscreen and back can fix this, but I couldn't get it to work yet.

lethal-guitar:
This changes the default window mode to windowed for Emscripten builds, fixing the problem where the game view was partially cut off in the browser window. Switching to fullscreen mode still does not do the right thing currently, leading to a cut off view and mouse position being offset in the options menu. But windowed mode works quite well.

In addition, we are now using the `--shell-file` option instead of a fully handwritten HTML file.

Fixes #569

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ Rigel Engine can be built to run on a web browser using [Emscripten](https://ems
 Note that this is still work in progress, as there are some [issues](https://github.com/lethal-guitar/RigelEngine/milestone/16) to resolve.
 
 Refer to Emscripten's [README](https://github.com/emscripten-core/emsdk) for instructions on how to install and activate Emscripten.
-You can verify the installation by running the `emcc` command. If it executes then everything is configured correctly.
+You can verify the installation by running the `emcc` command. If it executes without errors then everything is configured correctly.
 
 The instructions to build are the same as above, except for the CMake part:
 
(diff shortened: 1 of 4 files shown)


pratikone:
Oh wow, that was the reason for cutting off the game window. How did you figure it out ? And yay for figuring out on using shell-file. I had run circles around it earlier to figure out how to integrate it in cmake. There are more good compiler flags which can be passed for emscripten https://emscripten.org/docs/tools_reference/emcc.html#emcc-compiler-optimization-options  like safe_heap , sourcemap and others. Can be useful

lethal-guitar:
Experimentation: I remember you mentioned that switching to fullscreen and back could fix the cut off, but it didn't work for me. So then I thought maybe I'll try SDL's fullscreen switch via the options menu, and then I noticed that there was no cut off when in Windowed mode 🎉 

It's not clear to me yet why SDL's fullscreen doesn't toggle the browser's fullscreen.. I can see that there is [some code for it](https://hg.libsdl.org/SDL/file/a8f8265c328c/src/video/emscripten/SDL_emscriptenvideo.c#l320) in the Emscripten SDL backend.. 🤔

Issue #616, opened 2020-11-29
Consistently spell `Shareware` (not `ShareWare`)


lethal-guitar:

lethal-guitar:
fixed by #618

PR #617, opened 2020-12-01
Consistently spell Shareware (not ShareWare)

rbaliwal00:
consistent use of word shareware.

### DIFF
--- a/src/common/game_service_provider.hpp
+++ b/src/common/game_service_provider.hpp
@@ -43,7 +43,7 @@ struct IGameServiceProvider {
   virtual void stopMusic() = 0;
   virtual void scheduleGameQuit() = 0;
   virtual void switchGamePath(const std::filesystem::path& newGamePath) = 0;
-  virtual bool isShareWareVersion() const = 0;
+  virtual bool isSharewareVersion() const = 0;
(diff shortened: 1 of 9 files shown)


lethal-guitar:
Welcome to the project, @rbaliwal00, and thanks for your contribution! Appreciate it.

Before I merge, one question. When I added #616, I was also thinking of the function `isShareWareVersion` in the `IGameServiceProvider` class, which should be renamed to `isSharewareVersion`. Are you interested in doing that change as well? If so, I will leave this PR open so you can add another commit. But no problem if not, I'm happy to merge it as is also.

lethal-guitar:
@rbaliwal00 I see you've already added another commit, thank you! 👍🏻 Looks good!

Unfortunately, the build is failing now, as you can see [here](https://github.com/lethal-guitar/RigelEngine/pull/617/checks?check_run_id=1481390820#step:6:74) for example (you can get to that page by clicking the "Details" link next to the status checks above). It seems there are two more places in `menu_mode.cpp` that need to be changed.

Let me know if you'd like to fix those, if not, I'm happy to do it as well when I have time. And thanks again for your contribution! 👍🏻 

Closes #617

lethal-guitar:
closes #616

PR #633, opened 2021-01-05
FindFilesystem.cmake: Fix passing CMAKE_CXX_STANDARD to try_compile

peci1:
This is needed e.g. with gcc 8.3 which defaults to c++14. Without this fix, FindFilesystem.cmake fails to find any valid configuration.

See https://stackoverflow.com/a/47218035/1076564 .

### DIFF
--- a/cmake/Modules/FindFilesystem.cmake
+++ b/cmake/Modules/FindFilesystem.cmake
@@ -105,6 +105,13 @@ if(TARGET std::filesystem)
     return()
 endif()
 
+cmake_policy(PUSH)
+if(POLICY CMP0067)
+    # pass CMAKE_CXX_STANDARD to check_cxx_source_compiles()
+    # has to appear before including CheckCXXSourceCompiles module
+    cmake_policy(SET CMP0067 NEW)
+endif()
+
 include(CMakePushCheckState)
 include(CheckIncludeFileCXX)
 include(CheckCXXSourceCompiles)
@@ -229,3 +236,4 @@ if(Filesystem_FIND_REQUIRED AND NOT Filesystem_FOUND)
     message(FATAL_ERROR "Cannot Compile simple program using std::filesystem")
 endif()
 
+cmake_policy(POP)


lethal-guitar:
Hi @peci1, thanks for your PR! I'm actually not the author of `FindFileSystem.cmake` - I've copied it from [vector-of-bool/CMakeCM](https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake) (as noted here: https://github.com/lethal-guitar/RigelEngine/blob/ff280fdff192d4aac80ee777d71b8c1946c45957/cmake/Modules/FindFilesystem.cmake#L4) - so you might want to propose this change on the original repo as well, to make sure it gets fixed in the upstream version.

Regarding the fix, I haven't encountered this problem even when using GCC 8.3. Could it be that newer versions of CMake (I'm using 3.13) pass the flags correctly? RigelEngine currently requires at least CMake 3.13.

peci1:
I've looked at vector_of_bool, and their FindFilesystem.cmake contains `cmake_minimum_required(VERSION 3.10)` directly in the find module. This sets all policies from 3.10- to NEW, and CMP0067 was added in 3.8.

RigelEngine seems to specify cmake version only in the top-level cmakelists, and as you set 3.13, this is also not a problem for RigelEngine. So far so good.

However, as both your implementation and vector-of-bool's are referenced quite a lot in various forums, I guess it makes sense to try to keep the modules more or less self-contained. So one alternative to this PR would be to add `cmake_minimum_required(VERSION 3.13)`, but 3.13 is unncessarily high given 3.8 is sufficient. And it would be complicated to not forget to keep these versions in sync. That's why I suggest using the explicit policy setting.

lethal-guitar:
> This sets all policies from 3.10- to NEW, and CMP0067 was added in 3.8.
> as you set 3.13, this is also not a problem for RigelEngine

Ah, I see, I wasn't aware of this side effect when setting a minimum required version. That's good to know, thanks!

> That's why I suggest using the explicit policy setting.

Sounds reasonable. Thanks again for the PR!

PR #644, opened 2021-01-25
Removing implicit type conversion from int to size_t

pratikone:
VS was not compiling because of implicit conversion from int to size_t for these variables. I made them static_cast and the errors went away

### DIFF
--- a/src/renderer/renderer.cpp
+++ b/src/renderer/renderer.cpp
@@ -256,7 +256,7 @@ data::Image createWaterEffectColorMapImage() {
     pixels.push_back(loader::INGAME_PALETTE[index]);
   }
 
-  return data::Image{std::move(pixels), NUM_COLORS, NUM_ROWS};
+  return data::Image{std::move(pixels), static_cast<size_t>(NUM_COLORS), static_cast<size_t>(NUM_ROWS)};
 }
 
 
--- a/src/renderer/texture_atlas.cpp
+++ b/src/renderer/texture_atlas.cpp
@@ -74,7 +74,7 @@ TextureAtlas::TextureAtlas(
     throw std::runtime_error("Failed to build texture atlas");
   }
 
-  data::Image atlas{ATLAS_WIDTH, ATLAS_HEIGHT};
+  data::Image atlas{static_cast<size_t>(ATLAS_WIDTH), static_cast<size_t>(ATLAS_HEIGHT)};
 
   for (const auto& packedRect : rects) {
     atlas.insertImage(packedRect.x, packedRect.y, images[packedRect.id]);


lethal-guitar:
@pratikone thanks 👍🏻 Which VS version are you using? I don't get this warning on my machine but I haven't updated yet, I'm still on 16.7

I'm planning to change the constructor of `data::Image` to take `int` instead of `size_t`, btw - that's going to make a lot of annoying casts unnecessary

pratikone:
> 
> 
> @pratikone thanks 👍🏻 Which VS version are you using? I don't get this warning on my machine but I haven't updated yet, I'm still on 16.7
> 
> I'm planning to change the constructor of `data::Image` to take `int` instead of `size_t`, btw - that's going to make a lot of annoying casts unnecessary

that sounds good. yeah, I am on VS 2019 preview. I think it has some changes where it treats these warnings as errors. This came right up as an error in compilation and refused to be disabled until i casted it. 

Ok, sure. so if you are planning to change the constructor, we can close this PR without merging too since there is a better way planned

lethal-guitar:
> I am on VS 2019 preview

I see! I wonder if this might be a bug in that preview version of VS then? The variables are both `constexpr`, so I would expect the compiler to figure out that their values can be converted to `size_t` without information loss (since they are positive values) and thus no warning should be necessary. I'll investigate if this also happens with the latest non-preview version, and if not, then I'll think about making a bug report 🙂 

> Ok, sure. so if you are planning to change the constructor, we can close this PR without merging too since there is a better way planned

I'm planning to, but I'm actually not sure when I'll get around to it. So I'm happy to merge this in the meantime! Just want to make sure it's not a VS bug first.

Btw, did you also try running CMake with `-DWARNINGS_AS_ERRORS=OFF`? I enable warnings as errors by default in the CMake file, but the behavior can be overriden with that flag.

lethal-guitar:
@pratikone I just tried VS 2019 19.0 Preview 3, and I did not see the warning. So that would be an indication that it might be a bug in the version you're using. Which Preview version are you using exactly? And does the warning still occur if you update to the latest preview?

pratikone:
I updated my VS to a stable version  : 16.8.4 (VS 2019) and it still hits
10>D:\github\RigelEngine\src\renderer\texture_atlas.cpp(77): error C2398: Element '1': conversion from 'const int' to 'size_t' requires a narrowing conversion
10>D:\github\RigelEngine\src\renderer\texture_atlas.cpp(77): error C2398: Element '2': conversion from 'const int' to 'size_t' requires a narrowing conversion
10>D:\github\RigelEngine\src\renderer\renderer.cpp(259): error C2398: Element '2': conversion from 'const int' to 'size_t' requires a narrowing conversion
10>D:\github\RigelEngine\src\renderer\renderer.cpp(259): error C2398: Element '3': conversion from 'const int' to 'size_t' requires a narrowing conversion
10>Done building project "rigel_core.vcxproj" -- FAILED.

lethal-guitar:
@pratikone hm that's really strange, I'm not sure what could cause this difference, as I'm also using version 16.8.4.. Are you building a 32-bit build? 🤔 

Thanks for checking with the stable version, I'll merge this then.

PR #665, opened 2021-04-02
Replace boost program options library

### LINKED ISSUE #192: Replace boost::program_options
~~https://github.com/philsquared/Clara could be an option once it gets more stable.~~ Not maintained anymore.

I looked into https://github.com/jarro2783/cxxopts as well, but wasn't happy about compile times.

lethal-guitar:
Closes #192


lethal-guitar:
Currently blocked by https://github.com/bfgroup/Lyra/issues/49#issue-849259949

bencsikandrei:
Hey!

I'm curious, did you also consider: https://github.com/muellan/clipp?
I think it may be a better option.

lethal-guitar:
@bencsikandrei haven't seen that one yet, but indeed seems like it's not actively maintained. Might still be worth a look, if it works and has all the needed features it might be ok - not sure how well maintained Boost program options is, after all.

It seems surprisingly hard to find a good library that has all the needed features, is lightweight, and well maintained. Even though there's a ton of these - I guess a lot of people try to find a library, don't find one that makes them happy, and then write their own. 😄

adaugherity:
I'm sure you've got good reasons for not just using `getopt_long(3)` from libc, but that seems like the easiest way to remove any dependencies on large and/or potentially unmaintained option-parsing libraries... 😁

lethal-guitar:
> I'm sure you've got good reasons for not just using getopt_long(3) from libc

One reason is that it's not available on Windows/MSVC 🙂

PR #666, opened 2021-04-27
Use uint8 wraparound for RNG

bencsikandrei:
Hello,

Your CONTRIBUTING.md states that unsigned overflow is usable when useful.
Looks to me like the RNG is quite alright with a uint8_t since it has exactly 256 values (0-255).
Avoids a branch.

Do tell me if I've missed something and that IF has some other reason for being there.
I've also thought about adding a UT for the RNG (since I haven't found one). Just to make sure I didn't break anything. It would largely be doing a 1-1 check against the array tho..

Also, does it intentionally start at 1? i.e. that ++ before the access is there for a good reason right? It confused me a bit!<

Have a good one!

### DIFF
--- a/src/engine/random_number_generator.cpp
+++ b/src/engine/random_number_generator.cpp
@@ -16,6 +16,8 @@
 
 #include "random_number_generator.hpp"
 
+#include <limits>
+#include <type_traits>
 
 namespace rigel::engine {
 
@@ -47,6 +49,10 @@ const std::array<int, 256> RANDOM_NUMBER_TABLE{
 
 
 int RandomNumberGenerator::gen() {
+  static_assert(std::is_unsigned_v<decltype(mNextNumberIndex)>, "We rely on unsigned wraparound");
+  static_assert(RANDOM_NUMBER_TABLE.size() == static_cast<size_t>(std::numeric_limits<decltype(mNextNumberIndex)>::max()) + 1u,
+    "mNextNumberIndex must cover all values in the array"
+  );
   return RANDOM_NUMBER_TABLE[++mNextNumberIndex];
 }
 
--- a/src/engine/random_number_generator.cpp
+++ b/src/engine/random_number_generator.cpp
@@ -47,12 +47,7 @@ const std::array<int, 256> RANDOM_NUMBER_TABLE{
 
 
 int RandomNumberGenerator::gen() {
-  ++mNextNumberIndex;
-  if (mNextNumberIndex >= RANDOM_NUMBER_TABLE.size()) {
-    mNextNumberIndex = 0;
-  }
-
-  return RANDOM_NUMBER_TABLE[mNextNumberIndex];
+  return RANDOM_NUMBER_TABLE[++mNextNumberIndex];
 }
 
 }
--- a/test/test_rng.cpp
+++ b/test/test_rng.cpp
@@ -1,0 +1,54 @@
+/* Copyright (C) 2021, Nikolai Wuttke. All rights reserved.
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


lethal-guitar:
Thanks for your PR @bencsikandrei, and welcome to the project!

> Looks to me like the RNG is quite alright with a uint8_t since it has exactly 256 values (0-255).
Avoids a branch.

Indeed, that should be fine and makes sense. And it's less code and makes the `RandomNumberGenerator` object a bit smaller, so I think it's a good change! 👍🏻There's no particular reason for the `if`, that's just the way I wrote it.

> Also, does it intentionally start at 1? i.e. that ++ before the access is there for a good reason right?

Yes, that's very much intentional - we want to exactly reproduce the original binary's behavior, and it happens to do it this way for some reason. I agree that it's a bit odd 🙂 But there's a whole bunch of odd decisions in the original code 😄 

For reference, this is the assembly from the original DOS binary:

```
get_random_number proc far
inc     random_nr_index
mov     al, random_nr_index
mov     ah, 0
mov     bx, ax
mov     al, random_nr_table[bx]
retf
get_random_number endp
```

`random_nr_index` is a one byte global variable there. So your change actually makes it more similar to the original 🙂 

Regarding unit tests: Feel free to add one if you'd like! Testing all the values doesn't seem so useful, since as you mentioned it would basically just replicate the array. But testing that the wrap-around happens correctly could be useful.

I was also wondering if adding a `static_assert(RANDOM_NUMBER_TABLE.size() == std::numeric_limits<std::uint8_t>::max() + 1)` inside the `gen` function would be good? It's unlikely that someone would change the random table size without noticing the usage code right below, but it would be an additional piece of documentation making it explicit that we rely on the wrap-around.

bencsikandrei:
Hello, @lethal-guitar!

Thanks for the really nice and detailed reply!

I tried to make the best of what you described and what I had in mind and I think I struck a good balance. Added compile time checks for the RNG + a much slimmer unit test that only verifies what's not already checked at compile time (there's a bit of overlap, but it should be alright).

Funny thing, believe it or not, is I was also thinking about adding that assert you mentioned, just wasn't 100%. After the reply I had to add it in.

As always, if there's something else you feel needs adding I'd happily oblige!

Have a good one!

src/engine/random_number_generator.cpp
```diff
@@ -47,6 +49,10 @@ const std::array<int, 256> RANDOM_NUMBER_TABLE{
 
 
 int RandomNumberGenerator::gen() {
+  static_assert(std::is_unsigned_v<decltype(mNextNumberIndex)>, "We rely on unsigned wraparound");
```
lethal-guitar, line 52: 👍🏻

test/test_rng.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <catch.hpp>
+RIGEL_RESTORE_WARNINGS
+
+TEST_CASE("Random number generator works the same as the original") {
+  rigel::engine::RandomNumberGenerator rng;
+  SECTION("Values are generated in this particular order") {
+    SECTION("First generated value is 8") {
+      CHECK(rng.gen() == 8u);
+    }
+    SECTION("The period of the RNG is the original array's size") {
```
lethal-guitar, line 30: I think this is a good choice of what to test, period and start value makes sense. But I'd suggest making this test a bit stricter: If the period would somehow be different, the test would still pass as long as the 257th value is the same as the 1st value.

One way to make it stricter could be to generate `RANDOM_NUMBER_TABLE.size() * 2` values into a `std::vector`, and then check that the first 256 values are identical to the 2nd half of the vector.

What do you think?

lethal-guitar (commented):
Thanks for adding the tests! I've also done some semi-manual testing using the (still work in progress) demo playback mode, and I could see no difference between the previous version and yours.

I had one suggestion regarding the tests, let me know if you'd like to do it - I'm also happy to do it myself if you'd prefer moving on from this PR.

src/engine/random_number_generator.cpp
bencsikandrei: Hey, I'm on it!

test/test_rng.cpp
bencsikandrei: I think I covered what you described.
Was a fun test to write :)

lethal-guitar:
@bencsikandrei Thanks again! 👍🏻

PR #667, opened 2021-04-28
Add clang format

bencsikandrei:
Hello, again!

Saw you had this old issue with clang-format.
I've honestly tried to make it similar to what you had, but it's super impossible. I changed one flag stuff seemed to steer in the right direction then I want to emulate some other quirk in the existing formatting everything that worked before stops working ;p

Here's a first attempt anyhow.
I've also tried to do what you wanted in the comments, i.e. add some scripts to check formatting + a helper to format the whole project (src, test).

Note: I took the liberty to continue over the other MR I had, but if you want I can rebase.

Cheers!


lethal-guitar:
Oh wow, that's really awesome, thanks a lot for doing this! 🎉  I'll have a look at this as soon as I can!

lethal-guitar:
Thanks again for this, the scripts and CI action all LGTM 👍🏻 

> I've honestly tried to make it similar to what you had, but it's super impossible. I changed one flag stuff seemed to steer in the right direction then I want to emulate some other quirk in the existing formatting everything that worked before stops working ;p

Hehe I can imagine, really appreciate the effort! I think it's ultimately more valuable to have automated formatting than replicating the current formatting 100%, so all good.

I'll go through the diff and see if I find things that I'd like to tweak, and I'll also put some `clang-format: off` comments in a few select places (for example, the RNG table).

bencsikandrei:
Hey, you may also put those // clang-format off messages for the banners you print - cause those are super messed up by clang-format (https://github.com/lethal-guitar/RigelEngine/blob/ec1ff310dacdef8451f4cda32fc75577d3736552/src/main.cpp#L49)

Probably other similar things..

lethal-guitar:
I went through and adjusted the clang format file a bunch, and I think I've finally found a configuration that I like. I still need to go through the rest of the diff, only reviewed until `src/engine/physical_components.hpp` so far. So I might need to add a few more `clang-format off` comments here and there.

I also had to change the way the script invokes clang format to make it work on Windows. I assume it's due to some kind of length limitation, but basically when using `xargs` to pass the list of files to a single `clang-format` invocation, not all of the files were getting formatted. Using `find`/`exec` instead to spawn one `clang-format` per file catches all of the files.

In order to avoid duplication with the format checking script, I then changed the format check to run `format-all.sh` instead, and check that no files have changed.

lethal-guitar:
Found this GitHub action for checking if code is clang-formatted: https://github.com/DoozyX/clang-format-lint-action

Looks pretty nice, I'll try it out later today.

bencsikandrei:
> EDIT: Seems this needs a bit more work - the version of clang format I used locally is apparently newer than the one shipping with Ubuntu 20.04, so the job on CI doesn't recognize some of the flags. And unfortunately, the `find`/`exec` doesn't seem to report any non-zero exit codes from the executed commands, so the CI job is green even though it should fail with the last test commit I pushed..

Had the same kind of problems with other projects and a similar automation job.
What I ended up doing was using a docker image that has an exact version of clang-format (8 iirc) and used that across everything, from local formatting (in vscode) to the CI.

lethal-guitar:
> Also, between versions of clang-format, the behavior is different (unfortunately) - i.e. it actually formats differently using the same .clang-format file. We should pick a certain version and mention that in the Readme.

I noticed this as well when switching between computers that had different versions at first, really surprising/unexpected.. So yeah, we should definitely pick a version to be used. I've been using 12.0 for fine-tuning the config.
In addition to the README, I was thinking that the `format-all` script could also check the output of `clang-format --version` to make sure the right one is present.

Having a Docker container with the right version would be nice as well!

lethal-guitar:
I'm now fully done with going through the diff, and also extended the script to verify that the right version of clang-format is present. From my side, this is now basically ready to merge, only things remaining are adding some information about this to the contribution guide, and providing a Docker container. For the latter, I'm unsure if it would be better to add clang-format to the Ubuntu container that you recently added, or to have a separate Dockerfile for a container dedicated to running clang-format. Do you have any preferences in that regard, @bencsikandrei ?

bencsikandrei:
Hello!

(off-topic: been off for a while, but I'm still here)

Since that docker container is used for development it could integrate the clang-format tooling as well. But if you have settled for clang-format 12, not sure you can get that inside the 20.04 release of ubuntu (can you?).

One idea is to make a separate LLVM/Clang container (or a tooling container, w/e) that can be used standalone, but maybe also use it as a `FROM <...>` inside the development container for Ubuntu (make it a base of that image basically).
That does involve some housekeeping and, as you know, keeping things in multiple places can mean more build steps, more complex steps to follow, etc... but it may also be nice for the advanced user (?).

Long story short **I'd start by just having the simplest setup (i.e. with the same dev container for all tools, etc...)**

BUT your idea is also great - and for what it's worth, I think these steps are already pretty advanced - as in not many people are curious about them - so spicing them up a bit won't hurt ;p. Adding a version 12 for the project you sent seems trivial also. If the license permits it we could just take what's already done in there. Added benefit, you can test with llvm/clang (apple clang is not really clang anymore)

Cheers!

lethal-guitar:
Hey,

> (off-topic: been off for a while, but I'm still here)

all good 🙂 

> But if you have settled for clang-format 12, not sure you can get that inside the 20.04 release of ubuntu (can you?)

Unfortunately not, the latest one available in 20.04 is 11. Maybe I should try how the diff works out between 11 and 12.. 🤔 

Thanks for all the input! I'll think about it a bit more and maybe have a go at extending the Dockerfile or making another one. 👍🏻

lethal-guitar:
I had a look at using clang-format 11 instead of 12. The only real difference is in how it formats lambdas, and I actually kind of prefer it. Since version 11 is also available in Ubuntu 20.04, I've decided to make that version the required one, and then add it to the existing Dockerfile to keep things simple.

I'll update the documentation, and then merge this PR 🎉 Thanks again for getting things rolling on this, I'm really happy to have automated formatting now!

PR #668, opened 2021-04-28
Remove boost strings split usage

### LINKED ISSUE #662: Building without depending on libboost?
Hi there, @lethal-guitar !

On my continuing mission for native engines to try on SDL2/KMSDRM (SDL2 with no X11 depencencies!), I have just found this engine, and looks & sounds awesome on videos.

So I have tried to build it but seems to depend on boost, which is a HUGE lib!
Is there a way to build without depending on it? Checked CMakeLists.txt, but couldn't find a thing about it...

And, since it seems to run on OGA machines, I believe it must me possible to build without the heavy BOOST dependencies.

Thanks!

Manuel

EDIT: Even after installing libboost-dev, I still get this with `cmake -DUSE_GL_ES=ON -DCMAKE_BUILD_TYPE=Release ..`

```
manuel@hp15db0:~/src/RigelEngine/b4$ cmake -DUSE_GL_ES=ON -DCMAKE_BUILD_TYPE=Release ..
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake:117 (find_package):
  Could not find a package configuration file provided by
  "boost_program_options" (requested version 1.71.0) with any of the
  following names:

    boost_program_optionsConfig.cmake
    boost_program_options-config.cmake

  Add the installation prefix of "boost_program_options" to CMAKE_PREFIX_PATH
  or set "boost_program_options_DIR" to a directory containing one of the
  above files.  If "boost_program_options" provides a separate development
  package or SDK, be sure it has been installed.
Call Stack (most recent call first):
  /usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake:182 (boost_find_component)
  /usr/share/cmake-3.16/Modules/FindBoost.cmake:443 (find_package)
  CMakeLists.txt:53 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/manuel/src/RigelEngine/b4/CMakeFiles/CMakeOutput.log".
```

bencsikandrei:
Hello,

This one resolves #662, partially.
I've started with string `split` since I knew how to implement that.
Funny bit is that it ended up a bit faster than boost's version..

You should check out the benchmarks. I guess that's understandable since this version I added is much simpler and only uses a simple delimiter (one char).

But hey, I'll take it.

EDIT: add `trim[Right|Left]` and `startsWith`

Cheers!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ endif()
 
 option(USE_GL_ES "Use OpenGL ES instead of regular OpenGL" OFF)
 option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
-
+option(BUILD_BENCHMARKS "Build benchmarks" OFF)
 
 # Dependencies
 ###############################################################################
@@ -93,4 +93,8 @@ if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
 
     add_subdirectory(modding_tools)
     add_subdirectory(test)
+endif()
+
+if(BUILD_BENCHMARKS)
+    add_subdirectory(benchmark)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,8 +84,6 @@ endif()
 ###############################################################################
 
 add_subdirectory(3rd_party)
-
-rigel_configure_compiler_warnings()
 add_subdirectory(src)
 
 if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Emscripten")
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -1,0 +1,24 @@
+include(FetchContent)
+
+FetchContent_Declare(
+    benchmark
+    GIT_REPOSITORY https://github.com/google/benchmark.git
+    GIT_TAG        v1.5.3
+)
+FetchContent_GetProperties(benchmark)
+if(NOT benchmark_POPULATED)
+    set(BENCHMARK_ENABLE_GTEST_TESTS OFF)
+    set(BENCHMARK_ENABLE_INSTALL OFF)
+    set(BENCHMARK_ENABLE_TESTING OFF)
+    FetchContent_Populate(benchmark)
+    add_subdirectory(${benchmark_SOURCE_DIR} ${benchmark_BINARY_DIR})
+endif()
+
+add_executable(benchmarks
+    bench_string_utils.cpp
+)
+
+target_link_libraries(benchmarks PRIVATE
+    rigel_core
+    benchmark::benchmark_main
+)
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -22,3 +22,5 @@ target_link_libraries(benchmarks PRIVATE
     rigel_core
     benchmark::benchmark_main
 )
+
+rigel_enable_warnings(benchmarks)
--- a/benchmark/bench_string_utils.cpp
+++ b/benchmark/bench_string_utils.cpp
@@ -1,0 +1,146 @@
+/* Copyright (C) 2021, Nikolai Wuttke. All rights reserved.
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <benchmark/benchmark.h>
+
+#include <base/string_utils.hpp>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/algorithm.hpp>
+
+static void BMStringSplit(benchmark::State &state) {
+  std::string const kInputString = "Hello, world";
+  for (auto _ : state) {
+    const auto v = rigel::strings::split(kInputString, ',');
+    benchmark::DoNotOptimize(v);
+    benchmark::ClobberMemory();
+  }
+}
+
+BENCHMARK(BMStringSplit);
+
+static void BMBoostStringSplit(benchmark::State &state) {
+  const std::string kInputString = "Hello, world";
+  for (auto _ : state) {
+    state.PauseTiming();
+    std::vector<std::string> v;
+    state.ResumeTiming();
+    boost::split(v, kInputString, boost::is_any_of(","));
+    benchmark::DoNotOptimize(v);
+    benchmark::ClobberMemory();
+  }
+}
+
+BENCHMARK(BMBoostStringSplit);
+
+static void BMStartsWithTrueCase(benchmark::State &state) {
+  const std::string kInputString = "12341234";
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(
+      rigel::strings::startsWith(kInputString, "1234"));
+  }
+}
+
+BENCHMARK(BMStartsWithTrueCase);
+
+
+static void BMStartsWithTrueCaseLongString(benchmark::State &state) {
+  const std::string kInputString = "Iid7tUoNzQaGQjb9QqmuvqVQU9XbPmOvVbOI5ozuKdQN9bdHeP";
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(
+      rigel::strings::startsWith(kInputString, "Iid7tUoNzQaGQjb9QqmuvqVQU9Xb"));
+  }
+}
+
+BENCHMARK(BMStartsWithTrueCaseLongString);
+
+static void BMStartsWithFalseCase(benchmark::State &state) {
+  const std::string kInputString = "12341234";
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(
+      rigel::strings::startsWith(kInputString, "234"));
+  }
+}
+
+BENCHMARK(BMStartsWithFalseCase);
+
+
+static void BMBoostStartsWithTrueCase(benchmark::State &state) {
+  const std::string kInputString = "12341234";
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(
+      boost::starts_with(kInputString, "1234"));
+  }
+}
+
+BENCHMARK(BMBoostStartsWithTrueCase);
+
+static void BMBoostStartsWithFalseCase(benchmark::State &state) {
+  const std::string kInputString = "12341234";
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(
+      boost::starts_with(kInputString, "234"));
+  }
+}
+
+BENCHMARK(BMBoostStartsWithFalseCase);
+
+static void BMTrimLeft(benchmark::State &state) {
+  for (auto _ : state) {
+    state.PauseTiming();
+    std::string kInputString = "  1234  ";
+    state.ResumeTiming();
+    benchmark::DoNotOptimize(
+      rigel::strings::trimLeft(kInputString));
+  }
+}
+
+BENCHMARK(BMTrimLeft);
+
+static void BMTrimRight(benchmark::State &state) {
+  for (auto _ : state) {
+    state.PauseTiming();
+    std::string kInputString = "  1234  ";
+    state.ResumeTiming();
+    benchmark::DoNotOptimize(
+      rigel::strings::trimRight(kInputString));
+  }
+}
+
+BENCHMARK(BMTrimRight);
+
+static void BMTrim(benchmark::State &state) {
+  for (auto _ : state) {
+    state.PauseTiming();
+    std::string kInputString = "  1234  ";
+    state.ResumeTiming();
+    benchmark::DoNotOptimize(
+      rigel::strings::trim(kInputString));
+  }
+}
+
+BENCHMARK(BMTrim);
+
+static void BMBoostTrim(benchmark::State &state) {
+  for (auto _ : state) {
+    state.PauseTiming();
+    std::string kInputString = "  1234  ";
+    state.ResumeTiming();
+    boost::trim(kInputString);
+  }
+}
+
+BENCHMARK(BMBoostTrim);
--- a/src/base/string_utils.cpp
+++ b/src/base/string_utils.cpp
@@ -1,0 +1,23 @@
+#include "string_utils.hpp"
+
+#include <algorithm>
+#include <cassert>
+
+namespace rigel::strings {
+
+std::vector<std::string> split(std::string_view input, char delimiter) {
+  assert(static_cast<int>(delimiter) < 127 && "We only accept ASCII delimiters");
+
+  std::vector<std::string> output;
+  auto start = std::begin(input);
+  auto end = std::end(input);
+  decltype(start) next;
+  while ((next = std::find(start, end, delimiter)) != end) {
+    output.emplace_back(start, next);
+    start = std::next(next, 1);
+  }
+  output.emplace_back(start, next);
+  return output;
+}
+
+} // namespace rigel::strings
--- a/src/base/string_utils.cpp
+++ b/src/base/string_utils.cpp
@@ -20,4 +20,9 @@ std::vector<std::string> split(std::string_view input, char delimiter) {
   return output;
 }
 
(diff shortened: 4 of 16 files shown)


lethal-guitar:
Nice one, really appreciate it! 👍🏻 I'll check it out as soon as I can 🙂

src/common/string_utils.hpp
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace rigel::strings {
+
+/** Split a string based on a single char delimiter and put the result into output
+ */
+void split(std::vector<std::string> &output, std::string const &input,
```
lethal-guitar, line 26: I'd prefer returning the result, instead of using an out parameter. That would deviate more from the Boost API but I don't think that's an issue. Since there are not that many call sites it's easy to review the replacement.

Not sure why the Boost API was designed that way, maybe they didn't trust RVO back in the day? Or maybe it's for when you'd like to split multiple input strings into a single destination container? But that's not really a use case we have in RigelEngine right now, so we can go with a more straight-forward function signature.

test/test_string_utils.cpp
```diff
@@ -0,0 +1,47 @@
...
+    CHECK(v.size() == 2);
+    CHECK(v.front() == "hello");
+    CHECK(v.back() == " world");
+  }
+  SECTION("String with other white space delimiters") {
+    std::vector<std::string> v;
+    rigel::strings::split(v, "hello\nworld", '\n');
+    CHECK(v.size() == 2);
+  }
+}
```
lethal-guitar, line 47: Another edge case that could be worth testing is a string that ends with a delimiter, like `"hello\n".

test/test_string_utils.cpp
```diff
@@ -0,0 +1,47 @@
...
+  }
+  SECTION("String with no delimiter") {
+    std::vector<std::string> v;
+    rigel::strings::split(v, "hello", ' ');
+    CHECK(v.size() == 1);
+  }
+  SECTION("String with delimiter") {
+    std::vector<std::string> v;
+    rigel::strings::split(v, "hello, world", ',');
+    CHECK(v.size() == 2);
```
lethal-guitar, line 38: Suggestion for making the test assertions a bit nicer:

```cpp
const auto expected = std::vector<std::string>{"hello", " world"};
CHECK(v == expectedResult);
```

That way, both the length of the result as well as the values are checked in one statement.

You could also use deduction guides and literal postfix to make it even shorter:

```cpp
using namespace std::string_literals;
const auto expected = std::vector{"hello"s, " world"s};
```

Up to you which one you prefer, both work for me 🙂

src/common/string_utils.hpp
```diff
@@ -0,0 +1,29 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace rigel::strings {
```
lethal-guitar, line 22: I'd prefer to have this file in `src/base` instead of `src/common`. The former is where I mostly have general purpose utilities/standard library extensions, and string utilities fit in there quite well.

`common` on the other hand is more of a place for things that are needed to glue `frontend` and the rest of the code together - at least that's the idea, but in practice it mostly ended up being the place where I put things that don't really fit anywhere else 😅
 
Admittedly, that's not very clear from the name of the directory, I should think about a better name at some point..

src/common/string_utils.cpp
```diff
@@ -0,0 +1,19 @@
+#include "string_utils.hpp"
+
+#include <algorithm>
+
+namespace rigel::strings {
+
+void split(std::vector<std::string> &output, std::string const &input,
+           char delimiter) {
+  auto start = std::begin(input);
```
lethal-guitar, line 9: I wonder if we should `assert` that the delimiter is a plain ASCII character, i.e. less than 127? That way, string splitting can work correctly for UTF-8 encoded strings, without any surprises (as long as the delimiter is ASCII).

RigelEngine doesn't use Unicode anywhere so far, so it's not really a concern nowadays. But in the more distant future, I could imagine adding some localization features as an enhancement over the original game, so it would be a bit of future proofing.

What do you think?

lethal-guitar (commented):
Thanks a lot for helping out with the effort to get rid of Boost dependencies! And thanks for the comprehensive PR, I appreciate the tests & benchmarks and the corresponding README additions. 👍🏻 Nice to hear that the new version is faster. As you say, it's maybe not too surprising, with the Boost version being much more general.

I left a few comments regarding some details, but this already looks great overall, thanks again!

src/common/string_utils.hpp
bencsikandrei: I think they had to work with a million compilers, some of which we might have never even heard about. And those might not do RVO properly.
So happy you want to use RVO. From now on I'll definitely do that first. The only reason I did it like that was for the compatibility-ish..
Will do

test/test_string_utils.cpp
bencsikandrei: Sure thing, Have to admit I've not used Catch in a while!
CTAD is a thing of beauty, but somehow I prefer the first one (call me old fashion), and if both are OK with you I'll go with it.

src/common/string_utils.cpp
bencsikandrei: Will do.
Looks like something that can save some trouble in the long run, so I say why not.
Will be added along with the other changes.

bencsikandrei:
@lethal-guitar I've started looking at ways to replace 'startsWith' and 'trim' (and its flavors).

how do you want to proceed? do we do it in this same MR, or want me to open others for each of those? I'd really like to use the benchmarks here to see how much my changes differ from the boost:: functionality.

but hey, I can find a way

lethal-guitar:
@bencsikandrei cool! 👍🏻 I'd say feel free to add more commits to this PR, then we have all the Boost-related changes in one place

src/base/string_utils.cpp
```diff
@@ -20,4 +20,9 @@ std::vector<std::string> split(std::string const &input, char delimiter) {
   return output;
 }
 
+bool startsWith(std::string_view input, std::string_view prefix)
+{
+  return input.size() >= prefix.size() && input.substr(0, prefix.size()) == prefix;
```
lethal-guitar, line 25: This is simple and easy to read 👍🏻 so I would say it's fine if it's a little slower compared to the Boost version. We don't use it that frequently. And once we move to C++ 20, we can switch to the one from the standard lib (https://en.cppreference.com/w/cpp/string/basic_string/starts_with).

bencsikandrei: You gave me an awesome idea! Since C++20 is already partially supported by gcc/clang, I looked at how an implementation of this looks like and I made a small change that ended up being about 30% faster than the previous one, whilst remaining  readable (and extremely fast in debug mode - almost the same as in release actually).

lethal-guitar: Nice one! 👍🏻

lethal-guitar: Just out of curiosity, I had a quick look (https://github.com/boostorg/algorithm/blob/c9077bd49580bddfaa8a5c71abc4022a2b09d972/include/boost/algorithm/string/predicate.hpp) but I'm not really sure what makes it faster 🤔 it's more or less the same, except that they check both iterators against their respective range end on each iteration, instead of comparing the sizes upfront like you do. I would expect the latter to be faster since there should be fewer instructions per loop iteration, but who knows.. Could be interesting to compare the generated assembly.

In any case, your new version looks great to me 👍🏻

bencsikandrei:
If my search was correct that should be all of them. The only boost requirement you still have is the one you're fixing in #665

lethal-guitar:
> If my search was correct that should be all of them.

Awesome, I'll have a look!

benchmark/bench_string_utils.cpp
```diff
@@ -97,3 +97,42 @@ static void BMBoostStartsWithFalseCase(benchmark::State &state) {
 }
 
 BENCHMARK(BMBoostStartsWithFalseCase);
+
+static void BMTrimLeft(benchmark::State &state) {
+  std::string kInputString = "  1234 ";
```
lethal-guitar, line 102: I think you need to construct the string inside the benchmark loop, since the trimming mutates it in place - so after the first iteration, each subsequent trim call won't need to trim anything anymore.

We are going to also measure the string construction/destruction in that case, but assuming small string optimization, the time for that should be constant, so the comparison with the Boost implementation would still be valid I believe.

bencsikandrei: dang, how did I miss that?!

fixed now, plus, used benchmark's API to not time the string creation (dtor may still be timed tho)

lethal-guitar:
For future reference, benchmark results on my Windows Desktop:

```
Run on (8 X 3600 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 256 KiB (x8)
  L3 Unified 12288 KiB (x1)
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BMStringSplit                         126 ns          126 ns      4977778
BMBoostStringSplit                    718 ns          698 ns       896000
BMStartsWithTrueCase                 7.83 ns         7.95 ns    112000000
BMStartsWithTrueCaseLongString       8.69 ns         8.72 ns     89600000
BMStartsWithFalseCase                7.77 ns         7.85 ns     89600000
BMBoostStartsWithTrueCase            3.13 ns         3.15 ns    203636364
BMBoostStartsWithFalseCase           1.44 ns         1.44 ns    497777778
BMTrimLeft                            482 ns          547 ns      1000000
BMTrimRight                           480 ns          570 ns      1672533
BMTrim                                497 ns          594 ns      1000000
BMBoostTrim                           674 ns          767 ns       896000
```

lethal-guitar:
Hey @bencsikandrei, thanks again for this PR! I finally had the time to properly test everything, and it works great as far as I can tell. I made a few minor changes:

* Split adding the benchmarks into a dedicated commit
(more on why in a sec)
* Made `startsWith` `noexcept` as well - even though `std::equal` isn't marked as such, it should never throw in practice so that seemed fine to me. Let me know if you think that doesn't make sense!
* Made the trim unit tests a little more terse by inlining the expected value into the `CHECK` macro - this works fine for something like string literals, but looks a bit odd when doing it with `vector` or similar, so I usually prefer to have it outlined in a variable for the latter.
* Changed the CMake option name to `RIGEL_BUILD_BENCHMARKS`, as plural seemed a bit more fitting and it's also consistent with the option description that way.

Now this is basically ready to be merged from my side, except there's one little problem with the benchmarks.. I'm building with `-Werror` by default and on Mac/clang, that causes the build of googlebenchmark to fail due to warnings. I normally use `rigel_disable_warnings(target_name)` for 3rd party code, but it wasn't clear to me how I would apply that to the googlebenchmark target.

Maybe it would be better to not define `-Werror` globally, and instead have a `rigel_enable_warnings_as_errors()` helper function in CMake, and then apply that to all Rigel targets, instead of disabling warnings for 3rd party code?

Another thought I had was that we should probably build (even if we don't run them) the benchmarks on CI, as otherwise it might be too easy to accidentally break them for one of the platforms?

Let me know what you think, I'd be happy to merge this now without the benchmarks and add them back in a separate PR, or we can also sort it out as part of this PR - as you prefer 🙂

bencsikandrei:
Hey @lethal-guitar!

> * Split adding the benchmarks into a dedicated commit
>   (more on why in a sec)

Awesome work splitting it up. It's cleaner this way.

> * Made `startsWith` `noexcept` as well - even though `std::equal` isn't marked as such, it should never throw in practice so that seemed fine to me. Let me know if you think that doesn't make sense!
Good choice, my fingers were also itching to type that but I resisted because of equal not being noexcept. Glad you pulled the trigger. Even if it would throw, I think `std::terminate` is a good choice there, not like we're handling the exception anyway.

> * Changed the CMake option name to `RIGEL_BUILD_BENCHMARKS`, as plural seemed a bit more fitting and it's also consistent with the option description that way.
I agree!

> * Removed the reference to #662 from the commit message, since the issue won't be fully resolved until #665 has been merged.
Wasn't sure about this either, but good choice.


> Now this is basically ready to be merged from my side, except there's one little problem with the benchmarks.. I'm building with `-Werror` by default and on Mac/clang, that causes the build of googlebenchmark to fail due to warnings. I normally use `rigel_disable_warnings(target_name)` for 3rd party code, but it wasn't clear to me how I would apply that to the googlebenchmark target.

I think we can hack our way around this by setting some options on the `benchmark` target (maybe you already tried that?), but I  think I have a better way of handling it (it involves something that you wrote a few lines after).

> Maybe it would be better to not define `-Werror` globally, and instead have a `rigel_enable_warnings_as_errors()` helper function in CMake, and then apply that to all Rigel targets, instead of disabling warnings for 3rd party code? Then again, it would also be annoying to see warnings while the 3rd party targets are built.. 
^^ This.

> 
> Another thought I had was that we should probably build (even if we don't run them) the benchmarks on CI, as otherwise it might be too easy to accidentally break them for one of the platforms?

Agree.

> Let me know what you think, I'd be happy to merge this now without the benchmarks and add them back in a separate PR, or we can also sort it out as part of this PR - as you prefer
I think I can cook some cmake-foo rather quickly. I'll then activate the benchmarks in CI to see if it build.

CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ endif()
 
 option(USE_GL_ES "Use OpenGL ES instead of regular OpenGL" OFF)
 option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
-
+option(BUILD_BENCHMARKS "Build benchmarks" OFF)
```
bencsikandrei, line 21: Slight issue here, the flags don't agree. One's called BUILD_BENCHMARKS the other at line 98 checks for RIGEL_BUILD_BENCHMARKS

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    GIT_REPOSITORY https://github.com/google/benchmark.git
+    GIT_TAG        v1.5.3
+)
+FetchContent_GetProperties(benchmark)
+if(NOT benchmark_POPULATED)
+    set(BENCHMARK_ENABLE_GTEST_TESTS OFF)
+    set(BENCHMARK_ENABLE_INSTALL OFF)
+    set(BENCHMARK_ENABLE_TESTING OFF)
+    FetchContent_Populate(benchmark)
+    add_subdirectory(${benchmark_SOURCE_DIR} ${benchmark_BINARY_DIR})
```
bencsikandrei, line 14: we could do:
```
    # disable warnings for both targets inside the google `benchmark` build
    rigel_disable_warnings(benchmark)
    rigel_disable_warnings(benchmark_main)
```
but I would steer clear of this one since it's quite brittle. 
do we care if third party code has warnings when it builds itself? I would say no.
the better option is to have granular compiler warnings and set it ONLY on RigelEngine code, not on third party stuff.

CMakeLists.txt
lethal-guitar: Ah sorry, forgot to change the other one!

benchmark/CMakeLists.txt
lethal-guitar: Yeah that sounds good and cleaner/easier to maintain 👍🏻

bencsikandrei: These changes are present in the latest commit.
I'll resolve this since I started it and the changes are there.

lethal-guitar: Sure thing! 👍🏻 Thanks for making the changes!

lethal-guitar:
More benchmark results (2014 MacBook Pro):

```
Run on (4 X 2800 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 18.30, 11.13, 9.31
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BMStringSplit                         219 ns          218 ns      3202591
BMBoostStringSplit                   2337 ns         2344 ns       300370
BMStartsWithTrueCase                 3.98 ns         3.98 ns    174966756
BMStartsWithTrueCaseLongString       15.4 ns         15.3 ns     45158960
BMStartsWithFalseCase                2.94 ns         2.94 ns    236419394
BMBoostStartsWithTrueCase            2.47 ns         2.47 ns    281142563
BMBoostStartsWithFalseCase           1.22 ns         1.22 ns    564752961
BMTrimLeft                           1486 ns         1480 ns       471196
BMTrimRight                          1487 ns         1484 ns       468519
BMTrim                               1501 ns         1500 ns       465766
BMBoostTrim                          1592 ns         1590 ns       431398
```

lethal-guitar:
Aaaand it's merged. 🎉  Thanks again for this really nice contribution!!

PR #670, opened 2021-04-30
Add ubuntu latest dockerfile

bencsikandrei:
Dockerfile + instructions on how to build/run in docker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       with:
         submodules: recursive
     - name: Install dependencies
-      run: sudo apt-get update && sudo apt-get install libsdl2-dev libsdl2-mixer-dev libboost-dev libboost-program-options-dev g++-8
+      run: sudo ./docker/ubuntu-deps.sh
     - name: Run CMake (debug)
       run: CC=gcc-8 CXX=g++-8 cmake -H. -Bbuild_dbg -DCMAKE_BUILD_TYPE=Debug
     - name: Run CMake (release)
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+build
+.vscode
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,6 +8,7 @@
     * [Ubuntu 18.04](#linux-build-instructions-184)
     * [Fedora 31 or newer](#linux-build-instructions-fedora)
 * [Raspberry Pi builds](#raspi-build-instructions)
+* [Docker builds](#docker-build-instructions)
 * [OS X builds](#mac-build-instructions)
     * [Catalina (10.15) or newer](#mac-build-instructions-1015)
     * [Mojave (10.14) using clang 8](#mac-build-instructions-1014)
@@ -189,6 +190,37 @@ cd build
 cmake .. -DUSE_GL_ES=ON -DCMAKE_BUILD_TYPE=Release -DWARNINGS_AS_ERRORS=OFF
 make
 ```
+
+### <a name="docker-build-instructions">Docker builds</a>
+
+The provided docker image allows you to mimic one of the CI environments (ubuntu:latest).
+
+Build the image using
+
+```bash
+docker build . -f docker/ubuntu.Dockerfile -t rigel-build
+```
+
+Then run your container with:
+
+```bash
+# build it, debug it, etc..
+docker run -ti --cap-add=SYS_PTRACE --security-opt seccomp=unconfined -v $(pwd):/workdir -w /workdir rigel-build
+```
+
+And now follow the build steps from [Ubuntu](#linux-build-instructions-194).
+
+#### Running it inside the container
+
+You can also run the game from the container by following these instructions:
+
+1) on your host: `xhost +` (will allow connections from anything - you can be more granular if you wish)
+2) run the same build container with:
+```bash
+# ability to run the game with the GPU mapped + X forwarding
+docker run -ti --device=/dev/dri:/dev/dri --cap-add=SYS_PTRACE --security-opt seccomp=unconfined -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix --net=host -v $(pwd):/workdir -w /workdir rigel-build
+```
+3) build & run, same as you would on your local machine
 
 ### <a name="mac-build-instructions">OS X builds</a>
 
--- a/cmake/Modules/FindFilesystem.cmake
+++ b/cmake/Modules/FindFilesystem.cmake
@@ -217,9 +217,14 @@ if(CXX_FILESYSTEM_HAVE_FS)
         add_library(std::filesystem INTERFACE IMPORTED)
         target_compile_features(std::filesystem INTERFACE cxx_std_17)
         set(_found TRUE)
-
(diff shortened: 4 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       with:
         submodules: recursive
     - name: Install dependencies
-      run: sudo apt-get update && sudo apt-get install libsdl2-dev libsdl2-mixer-dev libboost-dev libboost-program-options-dev g++-8
+      run: CUSTOM_SUDO="sudo" ./docker/ubuntu-deps.sh
```
lethal-guitar, line 53: Would it work to run the script using something like `sudo bash ./docker/ubuntu-deps.sh`, and avoid the need for the `CUSTOM_SUDO` variable that way?

lethal-guitar (commented):
Thanks for the PR, looks good to me overall. Just wondering if the `CUSTOM_SUDO` workaround could be avoided somehow.

.github/workflows/ci.yml
bencsikandrei: Sure thing, unsure why I over-complicated it here. Was thinking about granularity. But it doesn't make much sense.

Fixed it now.

lethal-guitar:
I pushed some minor changes, and rebased. Looks ready to merge now for me 🙂 Did you still want to add/change anything, since the title still has `WIP`?

bencsikandrei:
Just adding some final touches - I think you'll like the small mods I'll add. Just need to test them. Give me a sec.

bencsikandrei:
@lethal-guitar 

I'll show you something cool that I found, let me know how you want to proceed!

Here goes:
I want to run the game from the container so I,
1) built it the same way as in the CI, with ubuntu-latest (which I assume means 20.04 - adding up to my theory is the fact that it has g++-8 directly in its repos - which 18.04 doesn't) - from the docker file you have already seen
1.1) run the container with `docker run -ti --device=/dev/dri:/dev/dri --cap-add=SYS_PTRACE --security-opt seccomp=unconfined -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix --net=host -v $(pwd):/workdir -w /workdir rigel-build`
2) compiled with the `CC=gcc-8 CXX=g++-8 cmake -H. -Bbuild_dbg -DCMAKE_BUILD_TYPE=Debug && cd build_dbg && make -j 4` commands
3) ran ./src/RigelEngine and got a SEGV
4) ran it through the debugger and got this BT:
```
#0  0x0000555555582e2f in std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::~vector() ()
#1  0x000055555558227a in std::filesystem::__cxx11::path::~path() ()
#2  0x00005555555866b0 in std::filesystem::__cxx11::path::_Cmpt::~_Cmpt() ()
#3  0x00005555555866cf in void std::_Destroy<std::filesystem::__cxx11::path::_Cmpt>(std::filesystem::__cxx11::path::_Cmpt*) ()
#4  0x0000555555585f42 in void std::_Destroy_aux<false>::__destroy<std::filesystem::__cxx11::path::_Cmpt*>(std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt*) ()
#5  0x00005555555851ed in void std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*>(std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt*) ()
--Type <RET> for more, q to quit, c to continue without paging--
#6  0x0000555555584485 in void std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt>(std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt*, std::allocator<std::filesystem::__cxx11::path::_Cmpt>&) ()
#7  0x0000555555582e45 in std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::~vector() ()
#8  0x000055555558227a in std::filesystem::__cxx11::path::~path() ()
#9  0x00005555555866b0 in std::filesystem::__cxx11::path::_Cmpt::~_Cmpt() ()
#10 0x00005555555866cf in void std::_Destroy<std::filesystem::__cxx11::path::_Cmpt>(std::filesystem::__cxx11::path::_Cmpt*) ()
#11 0x0000555555585f42 in void std::_Destroy_aux<false>::__destroy<std::filesystem::__cxx11::path::_Cmpt*>(std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt*) ()
#12 0x00005555555851ed in void std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*>(std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt*) ()
#13 0x0000555555584485 in void std::_Destroy<std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt>(std::filesystem::__cxx11::path::_Cmpt*, std::filesystem::__cxx11::path::_Cmpt*, std::allocator<std::filesystem::__cxx11::path::_Cmpt>&) ()
#14 0x0000555555582e45 in std::vector<std::filesystem::__cxx11::path::_Cmpt, std::allocator<std::filesystem::__cxx11::path::_Cmpt> >::~vector() ()
#15 0x000055555558227a in std::filesystem::__cxx11::path::~path() ()
#16 0x00005555555a0531 in rigel::createOrGetPreferencesPath[abi:cxx11]() ()
#17 0x00005555555a0894 in rigel::loadUserProfile() ()
#18 0x00005555555a0da3 in rigel::loadOrCreateUserProfile() ()
#19 0x00005555555811af in rigel::gameMain(rigel::CommandLineOptions const&) ()
#20 0x000055555558f115 in main ()
```
I know, it makes NO sense, it should not segfault here, but, bear with me: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90050

What ends up happening is that, I think, ubuntu 20.04 has a version of libstdc++ that contains the filesystem code (in the .so = version is 6.0.28 or something like that). Now, the CMake module you have picks that up since everything links just fine and says:
```
-- Looking for C++ include filesystem
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Success
```
It thus NEVER links with the mandatory libstdc++fs library (which has the ABI compatible code for fs in g++-8).

You would not have this problem with g++7 since that has only experimental fs (this another mangling).

This was fun ;p

I've solved it here, but if you wish I can take that bit out - we can merge that first, then do this one...

Cheers!

lethal-guitar:
@bencsikandrei oh wow, that's a nice find! I guess I've never tried running a build made with gcc 8, only tried compiling it (I'm using gcc 9 when working on Linux). I'll check it out, thanks a lot for the fix!

lethal-guitar:
Just tested it, I could reproduce the segfault and confirmed that your patch fixes it. Nice work, and thanks again! 👍🏻

Issue #419, opened 2019-08-09
Build (universal) binary packages for Linux


lethal-guitar:
Thanks to @bencsikandrei, we now have `.deb` packages.

In order to have a package that can run on a wider variety of distros, it would be cool to have one (or multiple) of AppImage, Snap or Flatpak as well.

Note: Flatpak has been started (see comments below).

Calinou:
I started working on a Flatpak: https://github.com/Calinou/io.github.RigelEngine

It builds and runs, but it can't find game data currently.

lethal-guitar:
@Calinou awesome, thanks a lot for your work! I'll test it as soon as I can.

> Ideally, it would be better if the user could drop data into a specified location in advance, and the game wouldn't ask for the game data path.

It's possible to specify the game path as command line argument when launching the game, and it won't ask for it in that case. Is there a way with Flatpak to include some kind of launch script maybe? That could make the fixed location work

Calinou:
> Is there a way with Flatpak to include some kind of launch script maybe? That could make the fixed location work

Sure :slightly_smiling_face: I can make it work out of the box then.

lethal-guitar:
We have an icon and a `.desktop` file now: https://github.com/lethal-guitar/RigelEngine/tree/master/dist/linux

Still working on the metainfo xml file.

Also worth noting, Boost is now an optional dependency, so the build could be simplified a bit now.

lethal-guitar:
Hey @Calinou, as of #852, we now have a metainfo xml file. I also continued working on the flatpak file, did some more testing, and I think it's ready now to be submitted to flathub. I'll make a PR shortly.

Thanks again for your work on getting this started!

lethal-guitar:
https://github.com/flathub/flathub/pull/3170

lethal-guitar:
Now published on flathub: https://flathub.org/apps/details/io.github.lethal_guitar.RigelEngine

pizzadude:
Nice! I wouldn't of heard of this if it weren't for you submitting it to flathub.

lethal-guitar:
@pizzadude happy to hear 🙂

PR #673, opened 2021-05-06
Add '.deb' package generation

bencsikandrei:
Hello,

Found this issue interesting: #419, gave it a shot with .deb packages (I knew how to do those). For the other I'd need some time to research what can be done.

Cheers!


.github/workflows/ci.yml
```diff
@@ -62,7 +62,13 @@ jobs:
     - name: Test (debug)
       run: cd build_dbg && ctest
     - name: Test (release)
-      run: cd build_dbg && ctest
+      run: cd build && ctest
+    - name: Package (deb)
+      run: cd build && cpack -G "DEB" ..
+    - uses: actions/upload-artifact@v2
+      with:
+        name: ubuntu_latest_build
```
bencsikandrei, line 70: I'm not fond of the `ubuntu-latest` tag

3rd_party/CMakeLists.txt
```diff
@@ -32,12 +32,13 @@ set(ENTITYX_DT_TYPE double CACHE INTERNAL "")
...
 
-add_subdirectory(dbopl)
-add_subdirectory(entityx)
-add_subdirectory(glad)
-add_subdirectory(speex_resampler)
-add_subdirectory(stb)
-add_subdirectory(imgui)
-add_subdirectory(imgui-filebrowser)
+# EXCLUDE_FROM_ALL to disable installing these HO or static library deps
+add_subdirectory(dbopl EXCLUDE_FROM_ALL)
```
bencsikandrei, line 36: since I added installation steps to the project, I wanted to make sure we don't install all the unnecessary development files (headers, static libs, etc...)

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
 
-project(RigelEngine)
+project(RigelEngine VERSION 0.7.1)
```
bencsikandrei, line 3: used the version you last tagged, but please let me know how you'd prefer to do the versioning @lethal-guitar

cmake/rigel_pack.cmake
```diff
@@ -0,0 +1,39 @@
...
+    # Debian
+    # (abencsik) intentionally left this here for future reference
+    # set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON) # automatically resolve deps
+    set(CPACK_DEBIAN_PACKAGE_DEPENDS "libc6 (>= 2.27), libstdc++6 (>= 8.4.0), libsdl2-2.0-0 (>= 2.0.4), libsdl2-mixer-2.0-0 (>= 2.0.1)")
+    set(CPACK_DEBIAN_PACKAGE_SECTION "games")
+    set(CPACK_DEBIAN_PACKAGE_RECOMMENDS "")
+    set(CPACK_DEBIAN_PACKAGE_SUGGESTS "")
+endif()
+
+include(CPack)
```
bencsikandrei, line 39: this file makes a .deb that looks like so:
```
new Debian package, version 2.0.
 size 1370640 bytes: control archive=422 bytes.
     361 bytes,    11 lines      control              
      66 bytes,     1 lines      md5sums              
 Architecture: amd64
 Depends: libc6 (>= 2.27), libstdc++6 (>= 8.4.0), libsdl2-2.0-0 (>= 2.0.4), libsdl2-mixer-2.0-0 (>= 2.0.1)
 Description: RigelEngine built using CMake
  A re-implementation of the game Duke Nukem II
 Maintainer: Nikolai Wuttke <lethal_guitar128@web.de>
 Package: rigelengine
 Priority: optional
 Section: games
 Version: 0.7.1
 Installed-Size: 3239
 ```

**Comment by bencsikandrei** [comment]
Tested the resulting artifact and it installs correctly and runs like a charm (on ubuntu latest that is ;p)

File: CMakeLists.txt (hunk 3 above)
Comment (lethal-guitar) [reply CMakeLists.txt:3]: Works for me! 👍🏻

File: 3rd_party/CMakeLists.txt (hunk 2 above)
Comment (lethal-guitar) [reply 3rd_party/CMakeLists.txt:36]: Makes sense - I remember running into this the last time I experimented with CPack, because the `entityx` library also has its own `install` steps that ended up adding some unneeded header files to the package.

I wasn't aware of this `EXCLUDE_FROM_ALL` flag, that's a very nice solution!

File: cmake/rigel_pack.cmake (hunk 4 above)
Comment (lethal-guitar) [reply cmake/rigel_pack.cmake:39]: Nice, looks good! Just the "RigelEngine built using CMake" part is a bit odd, maybe there's a way to override that?

File: .github/workflows/ci.yml (hunk 1 above)
Comment (lethal-guitar) [reply .github/workflows/ci.yml:70]: Maybe we should change this action to use the [ubuntu-18.04](https://github.com/actions/virtual-environments) environment instead of latest (assuming that gcc-8 can be installed without issues), or explicitly pin it to `20.04`. That seems better for CI stability in general. And then the artifact could be called accordingly.

**Review by lethal-guitar — COMMENTED** [review]
This is really cool, thanks a lot! And much simpler than I expected 😄 It all looks great to me as far as I can tell. I'll also test it on my Linux machine later today.

File: cmake/rigel_pack.cmake (hunk 4 above)
Comment (bencsikandrei) [reply cmake/rigel_pack.cmake:39]: Will fix it real quick.

**Comment by bencsikandrei** [comment]
Since g++ is the weapon of choice, how about we link libstdc++ statically?

This would make the binary bigger, BUT it would allow us to install on almost any linux distro since the rest of the deps are mostly the exact same across distros (they're mostly c libraries and are very stable).

It's not a silver bullet and it may have certain downsides, but it can work pretty well also.
All that we need is:
```
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
target_link_libraries(RigelEngine PRIVATE -static-libgcc -static-libstdc++)
endif()
```

lethal-guitar:
> Since g++ is the weapon of choice, how about we link libstdc++ statically?
>
> This would make the binary bigger, BUT it would allow us to install on almost any linux distro since the rest of the deps are mostly the exact same across distros

Sounds worth a shot to me! I like the idea of having a binary that runs on most distros, and maybe doing it this way already gets us there without needing the full on AppImage/Snap/Flatpak stuff.

Although I'd suggest adding an additional CMake option (something like `LINK_CPPSTDLIB_STATICALLY`), so that it's still possible to build with dynamic linking on g++ if desired. We can then set that option when building on CI.

lethal-guitar:
Just tested the artifact on my Ubuntu 20.04, and it worked without issues as well! 👍🏻 🎉 

I used the graphical installer (double-clicking the deb file), not the command line., and noticed that it says "License: Proprietary" - wonder if there's a way to tell CPack that it should display GPL-2 instead?

grafik

bencsikandrei:
I'll get back to this ASAP.
By the way, do you have an icon for Rigel? I may be able to make it show that in software center. Haven't done that before, but hey, might as well.

lethal-guitar:
> By the way, do you have an icon for Rigel? I may be able to make it show that in software center.

Not yet, but I was thinking the same - would be cool to have one for this 🙂 I've made an issue for now (#674)

bencsikandrei:
Hey,

Was trying real hard to make GPL sho up as a license. Didn't manage and found: 
https://stackoverflow.com/questions/50139467/open-source-deb-package-reported-as-proprietary
https://bugs.launchpad.net/ubuntu/+source/software-center/+bug/435183

Pretty old bug, but it seems it hasn't been fixed...

Unsure how to proceed really. What do you suggest?

lethal-guitar:
@bencsikandrei 

> Was trying real hard to make GPL sho up as a license. Didn't manage and found:

> Unsure how to proceed really. What do you suggest?

Oh wow, I didn't expect that it would be such an issue. Thanks a lot for investigating! I'd say it's fine to leave it as is - there's nothing we can really do about it, and it's not the end of the world either way.

.github/workflows/ci.yml
lethal-guitar: Maybe we should pin CI to 20.04 explicitly, then the artifact tag could be named accordingly?

(Re-added comment that disappeared)

CMakeLists.txt
lethal-guitar: Works for me 👍🏻 

(Re-added comment)

lethal-guitar:
Not sure what exactly happened, but somehow some of my comments have disappeared.. I've readded them 🤷🏻‍♂️

lethal-guitar:
@bencsikandrei thanks again for your work on this! 👍🏻 I was able to reuse a lot of the groundwork for making a Mac OS package as well (#696).

bencsikandrei:
Been a hectic couple of weeks... but I'm happy I could help!
I'll try to help with the #696, call me weird, but I like these kinds of issues

lethal-guitar:
@bencsikandrei no worries, that's how it is sometimes. Really appreciate all your contributions! 

Yeah if you have time, that'd be cool if you're interested in helping out with the OS X packaging. But no rush. In the meantime, I'll add some more info to the issue description so we have all the information in one place.

PR #677, opened 2021-05-21
Minor improvements to ArrayView

bencsikandrei:
* more constexpr
* more nodiscard
* remove unused headers

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -71,8 +71,8 @@ This module depends only on `base`, `data`, the standard library, and some 3rd p
 ### `renderer`
 
 Provides 2d-rendering functionality, using OpenGL as backend. This is meant
-to be kept fairly small in terms of the public intetface, and ideally general enough
-that other types of games could be built on top of it. The inner workings have
+to be kept fairly small in terms of the public interface, and ideally general enough
+that other types of games could be built on top of it.
 
 ### `sdl_utils`
 


test/test_array_view.cpp
```diff
@@ -0,0 +1,65 @@
...
+    static constexpr int arr[expectedSize]{0, 1, 2, 3, 4};
+    static constexpr rigel::base::ArrayView<int> arrv{&arr[0], 5};
+    static_assert(arrv.size() == 5);
+    static_assert(!arrv.empty());
+    static_assert(arrv[0] == 0);
+  }
+  SECTION("c array")
+  {
+    static constexpr auto expectedSize = 5;
+    static constexpr int arr[expectedSize]{0, 1, 2, 3, 4};
```
bencsikandrei, line 41: I'll add some more tests tomorrow

src/base/array_view.hpp
```diff
@@ -50,48 +48,43 @@ class ArrayView
...
     : mpData(array)
     , mSize(static_cast<size_type>(N))
   {
   }
 
   // implicit on purpose
-  ArrayView(const std::vector<T>& vec) noexcept // NOLINT
-    : mpData(vec.data())
-    , mSize(static_cast<size_type>(vec.size()))
+  template <typename Container>
```
bencsikandrei, line 66: this allows use with std::array, std::vector, static_vector, even string..

lethal-guitar: 👍🏻

test/test_array_view.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+TEST_CASE("Array view vector")
+{
+  const std::vector<int> vec{0, 1, 2, 3, 4};
+  const auto expectedSize = vec.size();
+  const rigel::base::ArrayView<int> arrv{vec};
+  CHECK(arrv.size() == expectedSize);
+  CHECK(!arrv.empty());
+  CHECK(*arrv.begin() == 0);
```
lethal-guitar, line 64: Maybe check for `std::equal(vec.begin(), vec.end(), arrv.begin())`?

test/test_array_view.cpp
```diff
@@ -0,0 +1,65 @@
...
+RIGEL_RESTORE_WARNINGS
+
+#include <array>
+#include <vector>
+
+TEST_CASE("Array view constexpr")
+{
+  SECTION("pointer and size")
+  {
+    static constexpr auto expectedSize = 5;
```
lethal-guitar, line 31: Do these need to be `static`, or would `constexpr` be sufficient?

src/base/array_view.hpp
```diff
@@ -50,48 +48,43 @@ class ArrayView
...
+  constexpr ArrayView(const Container& c) noexcept // NOLINT
+    : mpData(c.data())
+    , mSize(static_cast<size_type>(c.size()))
   {
   }
 
-  const_iterator begin() const { return mpData; }
+  constexpr const_iterator begin() const noexcept { return mpData; }
```
lethal-guitar, line 73: While we're at it, should we go all the way and make these also `[[nodiscard]]`? (And `front`, `back`, `data`, `operator[]`)?

lethal-guitar (commented):
Good changes, and thanks for adding tests, appreciate it!

test/test_array_view.cpp
bencsikandrei: added a bunch of tests, including ctad related ones

test/test_array_view.cpp
bencsikandrei: done

test/test_array_view.cpp
bencsikandrei: well, the expected size is ok being constexpr, the things requiring addresses-  like the arrays - need to also have a storage specifier. from what I understand, seems like the compiler needs to put them somewhere to actually do the computation

PR #748, opened 2021-09-29
Make Windows includes lowercase for MinGW

mnhauke:
MinGW-w64 ships all Windows SDK headers as lowercase, which prevents
cross-compiling this code from Linux.

Windows filesystems are case insensitive so it should work fine with
lowercase includes.


lethal-guitar:
Thank you! 👍🏻

PR #749, opened 2021-09-30
Fix openSUSE Tumbleweed build

lethal-guitar:
According to the [CMake docs](https://cmake.org/cmake/help/latest/module/FindThreads.html), `CMAKE_THREAD_LIBS_INIT` may be empty
even if threads are available. Therefore, we should use `Threads_FOUND`
instead. This fixes the build on openSUSE Tumbleweed.


lethal-guitar:
@mnhauke FYI!

mnhauke:
> @mnhauke FYI!

looks good, Thanks!

PR #771, opened 2021-11-04
Add option to run tests with ASan #244

bencsikandrei:
* cmake build type == Asan will add the required flags (fsanitizer and
-fno-omit-frame-pointer alongside the debug flags
* helpful messages are printed to see the build type
* usage with ctest is described in the BUILDING.md file

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -40,7 +40,7 @@ Each option can be either `ON` or `OFF`. The default is `OFF` for all options.
 
 * `USE_GL_ES`: Build against the OpenGL ES 2.0 API instead of OpenGL 3.0. See [OpenGL ES version](#linux-build-instructions-gles) for more info.
 * `WARNINGS_AS_ERRORS`: Make compiler warnings fail the build
-* `BUILD_TESTS`: Build the unit tests (`tests` target) 
+* `BUILD_TESTS`: Build the unit tests (`tests` target)
 * `BUILD_BENCHMARKS`: Build the benchmarks (`benchmarks` target)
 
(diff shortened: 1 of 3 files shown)


bencsikandrei:
@lethal-guitar I'd happily add options for the other sanitizers as well, if you wish!
but first, when you have some free time, let me know if this approach looks good to you.

Thanks!

lethal-guitar:
@bencsikandrei this is great, thanks a lot for this! 🎉 The approach seems perfect to me. If you'd like to add the other sanitizers, go for it, that'd be awesome!
I'm not familiar with the memcheck option of CTest, I will look into that.

bencsikandrei:
@lethal-guitar was looking into adding TSan and UBSan (MSan needs a bit more work) and I'm thinking why not activate multiple sanitizers at once?

TSan + ASan works without issues (ASan + MSan doesn't) but it would be cool to enable more at a time, to reduce the number of builds we need to make (cause in its current state for every santizer we have a new build with a new build type).

This approach seems to fall flat :( Although it really is nice for one sanitizer at a time.

I have an idea for multiple sanitizers also, something along these lines:
1) create a new `CMAKE_BUILD_TYPE` called `Sanitizer` (like the `Asan` thing) and use some other variables to determine which sanitizers to use in this build (like `RIGEL_ASAN`, `RIGEL_TSAN`, `RIGEL_UBSAN`, `RIGEL_MSAN`) - these could be additive
2) enable one or more of these alongside the `CMAKE_BUILD_TYPE` change

Do I do the one sanitizer at a time (like it is now just adding new `CMAKE_BUILD_TYPE`s) or expore the idea I had just now?

PR #779, opened 2021-11-14
Make intro movie files optional

### LINKED ISSUE #186: Make intro movie files optional for running RigelEngine
Currently, RigelEngine will terminate if it tries to show the intro movies and can't find the corresponding files (`NUKEM2.F1`, `NUKEM2.F2` etc.). Since the movies aren't mandatory for gameplay, it would be nicer to simply not show them if they aren't found. This would reduce the mandatory files required for playing to just `NUKEM2.CMP`.

All the relevant code can be found [here](https://github.com/lethal-guitar/RigelEngine/blob/f12a3c309dffada3e364991f51edffddab64fabf/src/frontend/intro_demo_loop_mode.cpp#L15). In that constructor, objects of type `ui::ApogeeLogo` and `ui::IntroMovie` are created. The constructors for these objects try to load the respective files, and throw an exception if they fail.

So one possible approach would be to surround the code that adds the movie objects with a `try`/`catch`.

PatriotRossii:
closes #186

### DIFF
--- a/src/frontend/intro_demo_loop_mode.cpp
+++ b/src/frontend/intro_demo_loop_mode.cpp
@@ -8,6 +8,8 @@
 #include "ui/duke_script_runner.hpp"
 #include "ui/menu_navigation.hpp"
 
+#include <stdexcept>
+
 
 namespace rigel
 {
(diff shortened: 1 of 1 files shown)


PatriotRossii:
I catch std::runtime_error cause load_file function throws and std::runtime_error exception

https://github.com/lethal-guitar/RigelEngine/blob/97732a910c19d06c680ee7b08e6ff1450b6a85e3/src/loader/file_utils.cpp#L43

lethal-guitar:
Nice one, thanks a lot! 👍🏻 Catching `std::exception` would also be an option, but it doesn't really matter in this case.

lethal-guitar:
Thanks again!

PR #783, opened 2021-11-17
Introduce a type alias Vec2f

### LINKED ISSUE #767: Introduce a type alias `Vec2f` and use it
Introduce a type alias for `Point<float>` called `Vec2f` and replace all uses of `Point<float>` with the new type alias.

If #766 is done first, then the alias would be for `Vex2T<float>`, of course.

PatriotRossii:
closes #767


lethal-guitar (approved):
Awesome, thanks a lot! 👍🏻

lethal-guitar:
I'll rebase this onto master when I get back to it

PR #812, opened 2022-01-08
Made Viewport consistent throughout

### LINKED ISSUE #795: Consistent spelling: `Viewport` instead of `ViewPort`
Spelling of the word "viewport" is inconsistent in the code, sometimes it's spelled `viewPort` and sometimes `viewport`. It would be nice to make it consistent.

mohitsaxenaknoldus:
Fixes #795


lethal-guitar:
Thanks for the PR, but I actually had something different in mind. #795 is about the names of identifiers in code, e.g. variables, parameters, function names etc. Where viewport is spelled as two camel-case words, i.e. `viewPort` or `ViewPort`, I'd like it to be spelled as `viewport` or `Viewport` instead. Otherwise, spelling or casing doesn't need to change, so e.g. `viewport` can stay as is.

Here is an example:

https://github.com/lethal-guitar/RigelEngine/blob/c1da6f88da0d676feef955629121f308f039df60/src/game_logic/camera.cpp#L115

The parameter `viewPortSize` should be called `viewportSize` instead in this case.

Hope this clarifies it, let me know if you have any questions!

mohitsaxenaknoldus:
> Thanks for the PR, but I actually had something different in mind. #795 is about the names of identifiers in code, e.g. variables, parameters, function names etc. Where viewport is spelled as two camel-case words, i.e. `viewPort` or `ViewPort`, I'd like it to be spelled as `viewport` or `Viewport` instead. Otherwise, spelling or casing doesn't need to change, so e.g. `viewport` can stay as is.
> 
> Here is an example:
> 
> https://github.com/lethal-guitar/RigelEngine/blob/c1da6f88da0d676feef955629121f308f039df60/src/game_logic/camera.cpp#L115
> 
> The parameter `viewPortSize` should be called `viewportSize` instead in this case.
> 
> Hope this clarifies it, let me know if you have any questions!

I am still confused about the requirements. Can you please explain again?

lethal-guitar:
> I am still confused about the requirements. Can you please explain again?

Sure. Ok, so let's look at the file [`src/game_logic/camera.cpp`](https://github.com/lethal-guitar/RigelEngine/blob/c1da6f88da0d676feef955629121f308f039df60/src/game_logic/camera.cpp#L115) as an example. In that file, on line 115, a parameter named `viewPortSize` is declared. This parameter should be named `viewportSize` instead.

The name `viewPortSize` in camel-case reads as "view port size", with "view" and "port" being two separate words due to the capitalization of "Port". But "viewport" is actually a single word, so the correct camel-case spelling for the words "viewport size" is `viewportSize`. 

This applies to *all* variables, function names etc. in the code. Whenever "viewport" is a part of the name, then the variable/function/... needs to be renamed so that "viewport" reads as a single word, by making the "P" in "port" lowercase. It doesn't matter what other words are part of the name, everything else remains unchanged, it's just about making the "P" lowercase whenever `viewPort` or `ViewPort` appears in the code 

Let's look at another example. In the file [`upscaling_utils.hpp`](https://github.com/lethal-guitar/RigelEngine/blob/c1da6f88da0d676feef955629121f308f039df60/src/renderer/upscaling_utils.hpp#L42), a struct named `WidescreenViewPortInfo` is declared. Again, we need to make "viewport" read as one word, so the struct should instead be called `WidescreenViewportInfo`.

mohitsaxenaknoldus:
> > I am still confused about the requirements. Can you please explain again?
> 
> Sure. Ok, so let's look at the file [`src/game_logic/camera.cpp`](https://github.com/lethal-guitar/RigelEngine/blob/c1da6f88da0d676feef955629121f308f039df60/src/game_logic/camera.cpp#L115) as an example. In that file, on line 115, a parameter named `viewPortSize` is declared. This parameter should be named `viewportSize` instead.
> 
> The name `viewPortSize` in camel-case reads as "view port size", with "view" and "port" being two separate words due to the capitalization of "Port". But "viewport" is actually a single word, so the correct camel-case spelling for the words "viewport size" is `viewportSize`. (I didn't know that "viewport" is a single word when I was writing a lot of this code).
> 
> This applies to _all_ variables, function names etc. in the code. Whenever "viewport" is a part of the name, then the variable/function/... needs to be renamed so that "viewport" reads as a single word, by making the "P" in "port" lowercase. It doesn't matter what other words are part of the name, everything else remains unchanged, it's just about making the "P" lowercase whenever `viewPort` or `ViewPort` appears in the code
> 
> Let's look at another example. In the file [`upscaling_utils.hpp`](https://github.com/lethal-guitar/RigelEngine/blob/c1da6f88da0d676feef955629121f308f039df60/src/renderer/upscaling_utils.hpp#L42), a struct named `WidescreenViewPortInfo` is declared. Again, we need to make "viewport" read as one word, so the struct should instead be called `WidescreenViewportInfo`.
> 
> And then the same for any other identifier in the code that is or contains `viewPort` (or `ViewPort`).

Please check now.

3rd_party/glad/3.0/include/glad/glad.h
```diff
@@ -952,8 +952,8 @@ typedef void (APIENTRYP PFNGLDEPTHRANGEPROC)(GLdouble n, GLdouble f);
 GLAPI PFNGLDEPTHRANGEPROC glad_glDepthRange;
 #define glDepthRange glad_glDepthRange
 typedef void (APIENTRYP PFNGLVIEWPORTPROC)(GLint x, GLint y, GLsizei width, GLsizei height);
-GLAPI PFNGLVIEWPORTPROC glad_glViewport;
-#define glViewport glad_glViewport
+GLAPI PFNGLVIEWPORTPROC glad_glviewport;
```
lethal-guitar, line 955: Anything in the `3rd_party` directory is library code, and therefore shouldn't be changed (unless when patching or modifying a library). Please revert all the changes in `3rd_party`.

Aside from that, `glViewport` is part of the OpenGL standard - it's an externally defined standard and not up to us to change it.

src/data/game_traits.hpp
```diff
@@ -38,29 +38,29 @@ struct GameTraits
...
+    viewportHeightPx};
 
   // The actual in-game viewport starts with an offset and is further reduced
   // to make room for the HUD. The right hand side features another 8px of
   // black border.
-  static constexpr base::Vec2 inGameViewPortOffset{8, 8};
-  static constexpr base::Extents inGameViewPortSize{
-    GameTraits::viewPortWidthPx - 16,
-    GameTraits::viewPortHeightPx - 8};
+  static constexpr base::Vec2 inGameviewportOffset{8, 8};
```
lethal-guitar, line 53: In this case, `ViewPort` had an uppercase `V` because it was in the middle of a longer name. The `V` needs to stay capitalized here. So the correct name should be `inGameViewportOffset`.

In general, the `V` in "viewport" can always stay unchanged. The issue is only about changing the `P` in `iewPort` to lowercase, it doesn't matter if it's `ViewPort` or `viewPort` - the "v" stays as it is, only the "P" needs to change

src/frontend/game.cpp
```diff
@@ -107,25 +107,25 @@ auto loadScripts(const loader::ResourceLoader& resources)
 }
 
 
-void setupRenderingViewport(
+void setupRenderingviewport(
```
lethal-guitar, line 110: This needs to remain an uppercase "V" here as well.

src/game_logic/game_world.hpp
```diff
@@ -115,13 +115,13 @@ class GameWorld : public entityx::Receiver<GameWorld>
   friend class rigel::GameRunner;
 
 private:
-  struct ViewportParams
+  struct viewportParams
```
lethal-guitar, line 118: Structs always start with an uppercase letter in the RigelEngine code base. So the correct name here is `ViewportParams`.

src/renderer/upscaling_utils.hpp
```diff
@@ -31,23 +31,23 @@ namespace rigel::renderer
...
+struct viewportInfo
 {
   base::Vec2 mOffset;
   base::Size<int> mSize;
   base::Vec2f mScale;
 };
 
 
-struct WidescreenViewPortInfo
+struct WidescreenviewportInfo
```
lethal-guitar, line 42: Here as well, the correct name would be `WidescreenViewportInfo`. I mentioned this already in my last response.

lethal-guitar (changes requested):
Thanks, this is going much more in the right direction. However, it's still not quite right. There are a lot of names that were changed correctly, but there are also many cases where you have changed the case of the letter "v" to lowercase. This is not necessary, only the "P" needs to change. The "V" should remain unchanged. In some cases, we want to have `viewport` with a lower-case "v", for example when a parameter or variable declaration starts with the word "viewport". In other cases, the code already has an uppercase "V". For example, when declaring a struct or class, or when the word viewport appears in the middle or end of an identifier name.

So, again: The case of the letter "v" should always stay the same!!

* `viewPort` changes to `viewport`
* `ViewPort` changes to `Viewport`
* `viewPortSomethingElse` changes to `viewportSomethingElse`
* `somethingViewPort` changes to `somethingViewport`
* `SomethingViewPort` changes to `SomethingViewport`

If you want to do this with a full-text search & replace, one way to do it would be to replace `iewPort` with `iewport` and disable matching complete words.

Are you familiar with the concept of camel case? I'd recommend reading about it, for example the [Wikipedia article](https://en.wikipedia.org/wiki/Camel_case). That might help understand what I'm going for with this issue.

lethal-guitar:
@mohitsaxenaknoldus I've edited my review comment above and added some more examples, hope this helps. Sorry that the issue description wasn't so clear, I'll try to give more examples next time.

PR #824, opened 2022-01-22
Create Shell script which runs clang-format on all staged files

### LINKED ISSUE #815: Add a `format-staged.sh` script
Currently, we have `tools/format-all.sh` to run clang-format on the code. Formatting all of the code can take a bit of time though. It would be handy to additionally have a 2nd script which runs clang-format only on files that are currently staged in git.

mohitsaxenaknoldus:
Closes #815

### DIFF
--- a/tools/format-staged.sh
+++ b/tools/format-staged.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+CLANG_FORMAT="clang-format-11"
+
+if ! which $CLANG_FORMAT >/dev/null 2>&1; then
+  CLANG_FORMAT="clang-format"
+fi
+
+if ! which $CLANG_FORMAT >/dev/null 2>&1; then
+  echo "error: clang-format not found, make sure it's in the PATH"
+  exit 1
+fi
+
+if ! $CLANG_FORMAT --version | grep "11\.\([0-9]\+\)\.\([0-9]\+\)" >/dev/null 2>&1; then
+  echo "error: wrong version of clang-format found, version 11.x.x needed"
+  exit 1
+fi
+
+STAGED_FILES="$(git diff --name-only --staged)"
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+
+for file in $STAGED_FILES; do
+  if [[ "$file" =~ \.[chi]pp$ ]]; then
+    $CLANG_FORMAT -i "$REPO_ROOT/$file"
+  fi
+done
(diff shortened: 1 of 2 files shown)


lethal-guitar:
Awesome, thanks! I'll check it out tomorrow.

tools/format-staged.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if ! $CLANG_FORMAT --version | grep "11\.\([0-9]\+\)\.\([0-9]\+\)" >/dev/null 2>&1; then
+  echo "error: wrong version of clang-format found, version 11.x.x needed"
+  exit 1
+fi
+
+staged_files=($(git diff --name-only --staged))
+
+for file in $staged_files; do
+  if [[ "$file" =~ \.cpp$ ]]; then
```
lethal-guitar, line 22: `.hpp` and `.ipp` are additional extensions that need to be included when formatting.

tools/format-staged.sh
```diff
@@ -0,0 +1,25 @@
...
+    $CLANG_FORMAT -i $file
```
lethal-guitar, line 23: Just to be on the safe side, I'd recommend quoting `$file` here, i.e. `"$file"`. In case there are files with spaces in the name or similar.

lethal-guitar (changes requested):
Thanks a lot for taking this on, really appreciate it! I've tested it and it generally works well, aside from not covering all file extensions (see comment). However, there are two cases that don't work currently:

1. When running the script from a path that's different from the repository root, the clang-format invocation fails with a "file not found error" due to the list of staged files being relative to the repository root. For example, running the script from the `build` directory: `../tools/format-staged.sh`. That's a very common use case, so it would be nice to make it possible.
2. Running the script via the Ubuntu Docker container doesn't work because there is no `git` installed in the container.

lethal-guitar (commented):
Thanks for the update, that should do it for the file extensions. Did you see my other comments regarding usage with Docker and from a different directory?

mohitsaxenaknoldus:
> Thanks for the update, that should do it for the file extensions. Did you see my other comments regarding usage with Docker and from a different directory?

Now you should be able to run it from any directory and I've added git to the Dockerfile because without it, it won't be possible to get the list of the changed files.

lethal-guitar:
Great, thanks! 👍🏻🎉 I'll be busy for the next couple days, but I should be able to get back to you sometime next weekend. Very happy to have this!

lethal-guitar (approved):

lethal-guitar:
Thanks again for your contribution. I had to make some more changes:

* Installing `git` in the Dockerfile didn't work because of a missing `-y` argument and usage of `sudo` (which isn't available in the Ubuntu base image)
* Only the first staged file was formatted, because the result of `git diff` wasn't escaped

I also simplified the file extension filtering.

PR #875, opened 2022-07-21
I replce every word in this code after i cloning it :)

GAbdulWahab:
* [x] I've compiled the code locally on my machine, and it builds without errors
 "There are problems because there something is wrong in the directory here "
Screenshot (1)

* [] I've launched the game on my machine after making my changes, and I can successfully start a new game from the main menu

* [x] If your changes are only touching documentation or similar files,
which have no influence on the behavior of the code,
you can ignore this of course.


lethal-guitar:
Nice, this looks good, and CI is green. I'll merge this later today.

Regarding the errors shown in your screenshot, this is because VSCode can't find the headers of the SDL library (which is a dependency of RigelEngine). I don't use VSCode myself, so I can't help you with setting it up unfortunately. I would suggest compiling the code directly in Visual Studio.

GAbdulWahab:
@lethal-guitar thank You for helping me sir <3

GAbdulWahab:
@lethal-guitar  ok, thank you sir <3

PR #927, opened 2025-04-18
Update JSON for Modern C++ to 3.12.0

tomkidd:
The version of JSON for Modern C++ in the code does something Xcode 16.3's bundled clang doesn't like (but 16.2 seems to be fine with it). Since the version in the code was 3.9.1, I updated it to 3.12.0.

* [ Y] I've compiled the code locally on my machine, and it builds without errors
* [ Y] I've launched the game on my machine after making my changes, and I can successfully start a new game from the main menu


lethal-guitar:
Thanks for the PR, appreciate it!

